PR #12, opened 2024-12-05
RDKB-58061 : simulated client is not associating to VAP

ramakanthtaraka:
Reason for change: Need to close the socket and create a new socket for each simulated client test case
Test Procedure: moving from private test cases to xfinity cases or viceversa check for client association and make sure client is associated to the vap.
Risks: None
Priority: P1


gsathish86:
@ramakanthtaraka Please sign your Contributor License Agreement.

gsathish86 (approved):

PR #13, opened 2024-12-05
platform: platform_pi.c: implement wifi_sendActionFrame

tuckerpo-cl:
Sends action frames via netlink.
-----------------------------------

Admittedly I am unsure if this is where this implementation belongs. Inspecting call sites in OneWiFi, it seems that `wifi_hal_send_mgmt_frame` would be the preferred forwarding call to make from `platform_pi.c` -- could anyone clarify?

i.e. in `platform_pi.c`:

```c
INT wifi_sendActionFrame(INT apIndex, mac_address_t MacAddr, UINT frequency, UCHAR *frame, UINT len)
{
    return wifi_hal_send_mgmt_frame(apIndex, MacAddr, frame, len, frequency);
}
```

as opposed to an in-line nl80211 impl.

### DIFF
--- a/platform/raspberry-pi/platform_pi.c
+++ b/platform/raspberry-pi/platform_pi.c
@@ -782,7 +782,8 @@ INT wifi_getApInterworkingServiceEnable(INT apIndex, BOOL *output_bool)
 
 INT wifi_sendActionFrame(INT apIndex, mac_address_t MacAddr, UINT frequency, UCHAR *frame, UINT len)
 {
-    return 0;
+    wifi_hal_send_mgmt_frame(apIndex, MacAddr, frame, len, frequency);
+    return RETURN_OK;
 }
 
 INT wifi_setDownStreamGroupAddress(INT apIndex, BOOL disabled)


platform/raspberry-pi/platform_pi.c
```diff
@@ -782,7 +782,62 @@ INT wifi_getApInterworkingServiceEnable(INT apIndex, BOOL *output_bool)
 
 INT wifi_sendActionFrame(INT apIndex, mac_address_t MacAddr, UINT frequency, UCHAR *frame, UINT len)
 {
-    return 0;
+    /*
```
amarnathhullur, line 785: Instead of fully implementing, please call wifi_hal_send_mgmt_frame in this function as you have suggested in the conversation.
i.e.
 return wifi_hal_send_mgmt_frame(apIndex, MacAddr, frame, len, frequency);

tuckerpo-cl: Thanks, I appreciate the clarification.

amarnathhullur (approved):

PR #15, opened 2024-12-09
Bpi onewifi integration

kpandu987:


src/wifi_hal_nl80211.c
```diff
@@ -5573,7 +5573,7 @@ static int phy_info_rates_get_hw_features(struct hostapd_hw_modes *mode, struct
 
 static int phy_info_handler(struct nl_msg *msg, void *arg)
 {
-    wifi_radio_info_t *radio;
+    wifi_radio_info_t *radio = (wifi_radio_info_t *) arg;
```
amarnathhullur, line 5576: This change should be related to singlephy, it should not be part of this commit

PR #42, opened 2025-01-24
Passed `freq` param through to `wifi_drv_send_mlme` function to allow for off-channel managment frame sending

bcarlson-dev:
Without this, the frequency parameter is unused, leading to a false impression of usefulness. The `freq` parameter can still be set to `0` to set it to the interface's frequency per the implementation of the `wifi_drv_send_mlme` function.


bcarlson-dev:
@narendradandu @amarnathhullur

gsathish86 (approved):

PR #61, opened 2025-02-06
RDKB-58155: Set MgmtPower for xfinity hotspot vaps

narendradandu:
Reason for change: Adding stub function for R-Pi
Test Procedure: NA
Priority: P1
Risks: None


narendradandu:
Already merged as part of #61

PR #81, opened 2025-02-25
Add kernel patches for Raspberry Pi

rakhilpe:
Add kernel patches for Raspberry Pi

Patch 1 - Address-the-crash-seen-while-sending-action-frame
Patch 2  - Added-MAC-ACL-support-for-RPI-by-modifying-kernel
Patch 3 - Modified-the-code-to-support-sending-all-kind-of-action-frame


mhughesacn:
Hi Rakhil:
This isn't a meta layer, so licensing is awkward.  Where did the the new source code in the two patches come from?  I am seeing a match to Broadcom code in the tool for one, but this is not indicated in the header.  The header indicates that all the modified code is original to you.
Also, how is the code in the patches licensed?  The Broadcom code match is ISC licensed but if there is any GPL code, then it will need to be hosted separately in its own directory.  An alternative would be to use a meta layer to host the patches.

PR #103, opened 2025-03-18
RDKBDEV-3009 : Fix memory leaks in "rdk-wifi-hal"

vevanga:
Reason for change: Fix memory leaks related to the usage of cJSON_Parse and cJSONPrint.
Test Procedure: Expected result - No memory leaks
Risks: None

Signed-off-by: venkatakrishna.vanga@cognizant.com

### DIFF
--- a/src/wifi_hal_dpp.c
+++ b/src/wifi_hal_dpp.c
@@ -716,11 +716,9 @@ void dpp_build_config(wifi_device_dpp_context_t *ctx, char* str)
         if (out)
         {
             printf("%s\n",out);
-
             /*let input string have json */
             strcpy(str, out);
             free(out);
-            out = NULL;
         }
 	
 	return;
--- a/src/wifi_hal_dpp.c
+++ b/src/wifi_hal_dpp.c
@@ -492,6 +492,8 @@ EC_POINT *dpp_build_point_from_connector_string(wifi_device_dpp_context_t *ctx, 
 
 	printf("%s:%d: point built\n", __func__, __LINE__);
 
+       cJSON_Delete(connector_json);
+
 	return instance->responder_connector;
 }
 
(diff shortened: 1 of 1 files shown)


pradeeptakdas:
Internal ticket created for tracking https://ccp.sys.comcast.net/browse/RDKCOM-5263

src/wifi_hal_dpp.c
```diff
@@ -711,10 +713,15 @@ void dpp_build_config(wifi_device_dpp_context_t *ctx, char* str)
...
+        if (out)
+        {
+            printf("%s\n",out);
 
-	/*let input string have json */
-	strcpy(str, out);
+            /*let input string have json */
+            strcpy(str, out);
+            free(out);
+            out = NULL;
```
vivianecordeiro-sky, line 723: this is a local variable, this line is unnecessary

vevanga: cJSON_Print() is calling  print() and inside it we are allocating memory dynamically and we are returning that pointer variable.Hence after copying the string to str, we are freeing the out variable.

vivianecordeiro-sky: your code fix is proper. I'm only asking you to remove the following line
 out = NULL;
THis line is not necessary since out is a local variable.

vevanga: Setting the out = NULL; // to prevent Coverity warning and also Dangling pointer problems.

vivianecordeiro-sky: Dangling pointer problems happen only the variable is external or global.
you should not see coverity issue if you remove this line
Please remove the line

vevanga: ok i will make the changes.( we can technically ignore setting out = NULL after calling free(ptr). However, it's generally considered good practice to set it to NULL)

gsathish86 (approved):

vivianecordeiro-sky (approved):

vivianecordeiro-sky:
Could you please rebase the code?

PR #126, opened 2025-04-02
Add kernel patches for Raspberry Pi

rakhilpe:
Add kernel patches for Raspberry Pi

Patch 1 - Address-the-crash-seen-while-sending-action-frame
Patch 2 - Added-MAC-ACL-support-for-RPI-by-modifying-kernel
Patch 3 - Modified-the-code-to-support-sending-all-kind-of-action-frame

PR #129, opened 2025-04-03
REFPLTB-3409: New NL80211 for ACL import changes causing AP issues in RPI

pganna856:
Reason for change:
    Added work around change for RPI.
Test Procedure:
    1. Bring up the device and check for SSIDs are UP or not using "iw dev" command.
    2. SSIDs should be broadcasted. Risks: None

### DIFF
--- a/src/wifi_hal_nl80211.c
+++ b/src/wifi_hal_nl80211.c
@@ -13554,7 +13554,7 @@ int wifi_drv_set_ap(void *priv, struct wpa_driver_ap_params *params)
         }
     }
 
-#if defined(NL80211_ACL) && !defined(PLATFORM_LINUX)
+#if defined(NL80211_ACL) && !defined(PLATFORM_LINUX) && !defined(_PLATFORM_RASPBERRYPI_)
     //Raspberry Pi kernel requires patching to support ACL functionality.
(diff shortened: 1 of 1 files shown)


cpokuru (commented):
As of now we are disabling ACL for opensource devices.we will get this back with proper testing.

amarnathhullur:
ACL will work with kernel changes, the patches for that are in this https://github.com/rdkcentral/rdk-wifi-hal/pull/126. As of now, allowing this change to go in to unblock RPI.  But the appropriate change would be to pull the kernel patches and rebuild the kernel. Discussing with Chandra, the patches as part of PR126 should be moved to meta-cmf-raspberrypi (https://code.rdkcentral.com/r/plugins/gitiles/rdk/components/generic/rdk-oe/meta-cmf-raspberrypi/+/refs/heads/rdk-next/recipes-kernel/linux/
recipes-kernel/linux - rdk/components/generic/rdk-oe/meta-cmf-raspberrypi) so that the kernel is patched with acl support. Once that is done, the current change should be reverted.

amarnathhullur (approved):

manigandanj:
Looks okay to me.

PR #143, opened 2025-04-14
RDKB-59269 Enable hostap_mgmt_frame_control

KarthikeyanNanjundan:
RDKB-59269 Enable hostap_mgmt_frame_control

Impacted Platforms:
All RDKB OneWiFi Platforms

Reason for change: Enable hostap_mgmt_frame_control by default.

Test Procedure: hostap_mgmt_frame_control should work fine and
client connectivity should be successful.

Risks: Low

Priority: P1

Signed-off-by: Karthikeyan Nanjundan Karthikeyan_Nanjundan@comcast.com


platform/broadcom/platform.c
```diff
@@ -3200,16 +3202,26 @@ static void platform_get_radio_caps_2g(wifi_radio_info_t *radio, wifi_interface_
     static const u8 ext_cap[] = { 0x85, 0x00, 0x08, 0x02, 0x01, 0x00, 0x40, 0x40, 0x00, 0x40,
         0x20 };
 #endif // XB10_PORT || SCXER10_PORT
+#if defined(TCHCBRV2_PORT)
+    static const u8 ext_cap[] = { 0x85, 0x00, 0x00, 0x02, 0x01, 0x00, 0x00, 0x40, 0x00, 0x00,
```
bogdanbogush2, line 3206: The driver already returns extended capabilities via NL: 0x85, 0x00, 0x08, 0x02, 0x01, 0x00, 0x00, 0x40, 0x00, 0x40, 0x20
The differences is in BSS transition capability. Could you check if is never enabled by Mesh?
The problem that if bit is 0 in extended_capa_mask we will not be able to enable feature by BSSTransitionActivated DM

src/wifi_hal_hostapd.c
```diff
@@ -1983,7 +1983,13 @@ int update_hostap_config_params(wifi_radio_info_t *radio)
 
     iconf->ht_rifs = radio->oper_param.band == WIFI_FREQUENCY_2_4_BAND;
 #endif
+
+#if defined(TCHCBRV2_PORT)
```
bogdanbogush2, line 1987: HAL has generic code, platform customization should be done via default settings, I suggest to have same behavior as for XB7/XB8 and enable this tag for 2.4G

KarthikeyanNanjundan: Removed the change to align with other platforms behaviour which use same driver.

KarthikeyanNanjundan: Removed the change to allign with other platforms behaviour which use same driver.

platform/broadcom/platform.c
KarthikeyanNanjundan: On discussion, decided to keep the extended capabilities align with CBRv2 driver ext capabilities.

bogdanbogush2 (approved):

vivianecordeiro-sky (approved):

gsathish86 (approved):

PR #145, opened 2025-04-15
REFPLTB-3426: Porting kernel patches to support ACL for RPI

pganna856:
Reason for change:
     Reverting the work around change added for ACL support.
Test Procedure:
    1. Bring up the device and check for SSIDs are UP or not using "iw dev" command.
    2. SSIDs should be broadcasted. 
Risks: None


anatar818 (approved):
Reverting the temporary fix

amarnathhullur (approved):

PR #175, opened 2025-04-30
RDKB-59604: avoid off-channel frame tx for DFS channel

bogdanbogush2:
Reason for change: frame tx fails on DFS channels

Test Procedure:
  - set DFS channel
  - check client can connect

Risks: Low

Priority: P1

### DIFF
--- a/src/wifi_hal_nl80211.c
+++ b/src/wifi_hal_nl80211.c
@@ -2966,6 +2966,7 @@ static int error_handler(struct sockaddr_nl *nla, struct nlmsgerr *err,
     int ack_len = sizeof(*nlh) + sizeof(int) + sizeof(*nlh);
 
     *ret = err->error;
+    wifi_hal_error_print("%s:%d: kernel error: %d\n", __func__, __LINE__, err->error);
 
     if (!(nlh->nlmsg_flags & NLM_F_ACK_TLVS))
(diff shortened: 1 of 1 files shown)


vivianecordeiro-sky (approved):

src/wifi_hal_nl80211.c
```diff
@@ -10978,7 +10975,7 @@ int wifi_drv_send_mlme(void *priv, const u8 *data,
 
     //wifi_hal_dbg_print("nl80211: send_mlme -> send_frame_cmd\n");
     res = nl80211_send_frame_cmd(interface, freq, wait, data, data_len,
-              use_cookie, noack, csa_offs, csa_offs_len);
+              use_cookie, 0, noack, csa_offs, csa_offs_len);
```
bcarlson-dev, line 10978: `wifi_hal_send_mgmt_frame` calls `wifi_drv_send_mlme` so just hardcoding 0, makes it so that existing methods that use `wifi_hal_send_mgmt_frame` cannot send off channel action frames. Please fix this issue by reusing the `offchanok` calculation made in  `wifi_drv_send_action`

bogdanbogush2: Thanks for spotting issue. I will fix in new PR.

PR #189, opened 2025-05-07
Remove dependency with VNTXER5_PORT for DT extender device

BijeeshaVSCom:
Reason for change:
Compilation support of TARGET_GEMINI7_2 flag

### DIFF
--- a/src/wifi_hal.c
+++ b/src/wifi_hal.c
@@ -502,7 +502,7 @@ INT wifi_hal_init()
         return RETURN_ERR;
     }
 
-#if defined(CONFIG_HW_CAPABILITIES) || defined(VNTXER5_PORT)
+#if defined(CONFIG_HW_CAPABILITIES) || defined(VNTXER5_PORT) || defined(TARGET_GEMINI7_2)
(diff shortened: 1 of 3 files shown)


rahulchandranr (approved):

src/wifi_hal.c
```diff
@@ -518,7 +518,7 @@ INT wifi_hal_init()
             interface = hash_map_get_next(radio->interface_map, interface);
         }
     }
-#endif // CONFIG_HW_CAPABILITIES || VNTXER5_PORT
+#endif // CONFIG_HW_CAPABILITIES || VNTXER5_PORT || TARGET_GEMINI7_2
```
gsathish86, line 521: Please check.  You need to add TARGET_GEMINI7_2 macro in if defined as well.  It is missing.

gsathish86 (changes requested):
Please address the comments

src/wifi_hal.c
BijeeshaVSCom: This is a modification in the comment added at #endif

gsathish86 (approved):

PR #228, opened 2025-06-10
RDKBWIFI-39: Allow wildcard station scan

bcarlson-dev:
Reason for change:

Currently, there is no way to get all of the results of a scan if you are scanning as a station. This allows you to get all of those results by just providing an empty (wildcard) SSID.

Test Procedure:

Call `wifi_hal_startScan` (through some means) on a station VAP with an empty SSID and see that all scan results appear. Also viewable through `/tmp/wifiHalStats`

Priority: P1
Risks: Low

Related to: 
- https://github.com/rdkcentral/OneWifi/pull/386
- https://github.com/rdkcentral/OneWifi/pull/387


bcarlson-dev:
@narendradandu

gsathish86 (approved):

PR #238, opened 2025-06-18
DTMESH-461 [DT] Changes for Onewifi Leader mode bring-up

MageshRenganathan:
Reason for change: Backhaul AP creation fails when bridge name is null which is not mandatory for all vaps

### DIFF
--- a/src/wifi_hal_rdk_util.c
+++ b/src/wifi_hal_rdk_util.c
@@ -361,10 +361,6 @@ int validate_wifi_interface_vap_info_params(wifi_vap_info_t *vap_info, char *msg
         ret = RETURN_ERR;
         snprintf(msg + strlen(msg), len - strlen(msg), " showSsid: %d", bss_info->showSsid);
     }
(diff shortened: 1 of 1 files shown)


src/wifi_hal_rdk_util.c
```diff
@@ -362,7 +362,7 @@ int validate_wifi_interface_vap_info_params(wifi_vap_info_t *vap_info, char *msg
         snprintf(msg + strlen(msg), len - strlen(msg), " showSsid: %d", bss_info->showSsid);
     }
     if (strncmp(vap_info->bridge_name, "", strlen(vap_info->bridge_name)) == 0) {
```
vivianecordeiro-sky, line 364: remove all if code

vivianecordeiro-sky (changes requested):

vivianecordeiro-sky (approved):

gsathish86 (approved):

PR #243, opened 2025-06-20
XER10-1812 : bssehtmode does not match eht

jonathanwu-csv:
Reason for change: bssehtmode is 1 when EHT is disabled.
Test Procedure: Enable or disable EHT and BSS bssehtmode should be same as EHT.
Risks: None
Priority: P2

### DIFF
--- a/platform/broadcom/platform.c
+++ b/platform/broadcom/platform.c
@@ -3566,6 +3566,7 @@ static void platform_set_eht(wifi_radio_index_t index, bool enable)
 {
     bool eht_enabled;
     bool radio_up;
+    int bss;
 
     eht_enabled = platform_is_eht_enabled(index);
     if (eht_enabled == enable) {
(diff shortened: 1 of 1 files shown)


platform/broadcom/platform.c
```diff
@@ -3577,6 +3578,10 @@ static void platform_set_eht(wifi_radio_index_t index, bool enable)
         v_secure_system("wl -i wl%d down", index);
     }
     v_secure_system("wl -i wl%d eht %d", index, (enable) ? 1 : 0);
+    v_secure_system("wl -i wl%d eht bssehtmode %d", index, (enable) ? 1 : 0);
+    for (bss = 1; bss <= 7; bss++) {
+        v_secure_system("wl -i wl%d.%d eht bssehtmode %d", index, bss, (enable) ? 1 : 0);
+    }
```
gsathish86, line 3584: If you are going to use it only for XER10, please have this under this port flag SCXER10_PORT.

gsathish86 (approved):

PR #245, opened 2025-06-23
RDKB-60218 & RDKB-60219

BijeeshaVSCom:
[RDKB-60218] Incorrect MAX channel set for 2G band in rdk-wifi-hal
[RDKB-60219] [XLE][DT] Backahul station disconnect during Ap push when backahul connected through non default channels


src/wifi_hal_nl80211_events.c
```diff
@@ -628,12 +628,20 @@ static void nl80211_connect_event(wifi_interface_info_t *interface, struct nlatt
     mac_addr_str_t bssid_str;
     wifi_bss_info_t *backhaul;
     wifi_vap_security_t *sec;
+    wifi_vap_info_t *vap;
```
gsathish86, line 631: Please remove this variable as it is not used inside the function

src/wifi_hal_nl80211_events.c
```diff
@@ -628,12 +628,20 @@ static void nl80211_connect_event(wifi_interface_info_t *interface, struct nlatt
...
+    wifi_radio_operationParam_t *radio_param;
+
     sec = &interface->vap_info.u.sta_info.security;
 
     backhaul = &interface->u.sta.backhaul;
 
     wifi_hal_dbg_print("%s:%d:bssid:%s frequency:%d ssid:%s\n", __func__, __LINE__,
         to_mac_str(backhaul->bssid, bssid_str), backhaul->freq, backhaul->ssid);
+    vap = &interface->vap_info;
+    radio = get_radio_by_rdk_index(vap->radio_index);
```
gsathish86, line 642: Instead of vap, interface->vap_info is used directly from arg.  I don't see vap is used anywhere apart from this place.

gsathish86 (changes requested):
Please address the comments.

src/wifi_hal_nl80211_events.c
BijeeshaVSCom: Removed vap and used interface->vap_infob directly in the code

PR #308, opened 2025-08-14
RDKB-60953, BCOMB-3031: Add AMSDU TID handling

bmilcz-comcast:
Reason for change: Implements handling of AMSDU TID
Test Procedure: Verify build, set/get AMSDU TID
Risks: None
Priority: P2

this PR depends on https://github.com/rdkcentral/rdkb-halif-wifi/pull/45


src/wifi_hal_nl80211.c
```diff
@@ -7429,6 +7429,47 @@ int nl80211_update_wiphy(wifi_radio_info_t *radio)
...
+int nl80211_set_amsdu_tid(wifi_interface_info_t *interface, uint8_t *amsdu_tid)
+{
+    wifi_hal_dbg_print("%s:%d: Setting AMSDU for interface->name=%s\n", __func__, __LINE__,
+        interface->name);
+
+    struct nl_msg *msg;
+    struct nlattr *nlattr_vendor = NULL;
+
+    // Create the vendor-specific command message
+    msg = nl80211_drv_vendor_cmd_msg(g_wifi_hal.nl80211_id, interface, 0, OUI_COMCAST,
```
amarnathhullur, line 7441: If there is a driver which doesn't support this vendor command then the error would print and return -1 from this function. Is there any side effect of this? 
Essentially, want to confirm that there is no impact in operation for drivers not supporting this vendor command.

bmilcz-comcast: So on driver side - if the command RDK_VENDOR_NL80211_SUBCMD_SET_AMSDU_CONFIG is not implemented there should be no impact. Driver essentially should ignore the commands it does not know.

As for this specific line failing - this would be a problem on netlink library API level, thus would not relate to the command itself or it's implementation. Currently, as for similar commands and as you have noted, -1 would be returned with error line printed out, but no other action would be undertaken (see wifi_hal.c).

Side effect would be a desync in datamodel values with regard to AMSDU setting for specific Traffic IDs which were changed.

bmilcz-comcast: Update: Talked with @narendradandu and the function is too important to fail at trivial parameter set - we just signal the error via print and that is it.

narendradandu (approved):

PR #326, opened 2025-08-28
RDKCOM-5421: RDKBDEV-3279 Need to align stub API implementations in intel/platform.c file in rdk-wifi-hal.

aprasad-97:
Reason for change: I can see that new APIs are used in RDKB25Q2 code, but their definitions are missing in intel/platform.c file. Due to this, we faced compile issue during build in MXL setup. So, we will bring these API definitions in intel/platform.c and fix this issue.

Risks: Low
Signed-off-by: Aiswarya Prasad <aprasad@maxlinear.com>

### DIFF
--- a/.github/community_authors.json
+++ b/.github/community_authors.json
@@ -1,0 +1,22 @@
+{
+  "community_users": [
+    "amarnathhullur",
+    "rakhilpe",
+    "pratheepa1995",
+    "prashant-singh1",
+    "pradeeptakdas",
+    "pganna856",
+    "Nikita-Hakai",
+    "tuckerpo-cl",
+    "sriraaman-c",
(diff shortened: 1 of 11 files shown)


platform/intel/platform.c
```diff
@@ -1636,3 +1648,40 @@ int platform_get_radio_caps(wifi_radio_index_t index)
 {
     return 0;
 }
+
+#ifdef CONFIG_IEEE80211BE
```
narendradandu: CONFIG_IEEE80211BE is not enabled for intel platforms, can you please describe more on what errors are seen

aprasad-97:
> Hi @aprasad-97
> 
> Can you please fix the review comments suggested by Narendra?

Addressed the comments @pradeeptakdas

PR #333, opened 2025-09-01
Wifi7 bring up on DT platform

MG-kildon:


platform/qualcomm/platform_ext.c
```diff
@@ -1183,6 +1203,89 @@ static int qca_add_intf_to_bridge(wifi_interface_info_t *interface, bool is_mld)
...
+    char mld_ifname[32];
+    char ifname[32];
+    char mld_mac_addr[ETH_ALEN];
+    char mld_mac_str[18];
+
+    if (interface == NULL || radio == NULL) {
+        wifi_hal_error_print("%s:%d: Invalid arguments\n", __func__, __LINE__);
+        return RETURN_ERR;
+    }
+    if ((int)interface->vap_info.vap_index >= 0) {
```
vivianecordeiro-sky, line 1258: this check does not make sense since the variable is unsigned and get_interface_name_from_vap_index will check if value is valid

PR #352, opened 2025-09-09
RDKBACCL-1067 : bpi default ssid names got changed in latest tip code

kpandu987:
Reason for change: Updated bpi default ssid name in platform layer
Test Procedure: Compilation and build checked.
Risks: Low


amarnathhullur (approved):
Approved

PR #568, opened 2026-02-13
RDKCOM-5529: RDKBDEV-3367 Disable EasyMesh services runtime based on flag

satish-kotapati:
Reason for change: provision to handle easymesh based on flag

Test Procedure: OneWifi running while EasyMesh services are disabled
Risks: None


pradeeptakdas:
Thanks @satish-kotapati for confirming that you have fixed  the review comments.
We will initiate the review with component owner

narendradandu:
Can we better understand the use case for runtime EasyMesh disable?

If the requirement is to operate without EasyMesh, wouldn't it be cleaner to build with EasyMesh components excluded and OneWifi remains independent rather than introducing an NVRAM flag and additional runtime checks? This appears to be addressing a dependency on EasyMesh specific interfaces rather than a core OneWifi requirement.

narendradandu (approved):

PR #583, opened 2026-02-23
RDKB-61540: MLO link reconfiguration for BPi-R4

bmilcz-comcast:
Reason for change:Enable possibility of dynamically enabling/disabling MLO links for multiple MLD devices.
Test Procedure: Verify build, test MLO VAPs, enable/disable links in MLO VAPs, check if appropriate singular VAP gets reset properly, confirm connectivity with a client device.
Risks: None
Priority: P2

### DIFF
--- a/platform/banana-pi/platform.c
+++ b/platform/banana-pi/platform.c
@@ -25,11 +25,12 @@
 * Licensed under the BSD-3 License
 **************************************************************************/
 
-#include <stddef.h>
-#include <string.h>
-#include <stdlib.h>
-#include "wifi_hal_priv.h"
 #include "wifi_hal.h"
+#include "wifi_hal_priv.h"
+#include <net/if.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
 
 #define NULL_CHAR '\0'
 #define NEW_LINE '\n'
@@ -44,6 +45,7 @@
 
 int wifi_nvram_defaultRead(char *in,char *out);
 int _syscmd(char *cmd, char *retBuf, int retBufSize);
+int dealloc_mld(wifi_interface_info_t *interface);
 
 typedef struct {
     mac_address_t *macs;
@@ -150,7 +152,7 @@ int platform_create_vap(wifi_radio_index_t index, wifi_vap_info_map_t *map)
             continue;
         }
 
-        if (!wifi_hal_is_mld_enabled(interface)) {
+        if (!wifi_hal_is_mld_enabled(interface) || !vap->u.bss_info.enabled) {
             continue;
         }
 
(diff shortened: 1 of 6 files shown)


bmilcz-comcast:
recheck

src/wifi_hal_nl80211.c
```diff
@@ -15851,10 +15854,31 @@ int nl80211_register_bss_frames(wifi_interface_info_t *interface)
 }
 #endif
 
+void nl80211_unregister_spurious_frames(wifi_interface_info_t *interface)
+{
+    if (interface->spurious_frames_registered == 0) {
+        wifi_hal_dbg_print("%s:%d: interface:%s spurious frames not registered\n", __func__,
+            __LINE__, wifi_hal_get_interface_name(interface));
```
mateuszCieslak-GL: keep the convention ;) 

```suggestion
            __LINE__, interface->name);
```

src/wifi_hal_nl80211_utils.c
```diff
@@ -5966,3 +5998,181 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+    }
+#endif
+    pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+
+    /* Prevent hostap calling set_ap when client is removed due to VAP disable before
+     * start_bss */
+    interface->in_reconf = true;
+
+    wifi_hal_info_print("%s:%d: interface:%s disable AP\n", __func__, __LINE__, interface_name);
+    nl80211_enable_ap(interface, false);
```
mateuszCieslak-GL: it would be nice to catch error here, even under (if (unlikely( ... < 0) )..

platform/banana-pi/platform.c
```diff
@@ -321,33 +327,214 @@ int platform_pre_create_vap(wifi_radio_index_t index, wifi_vap_info_map_t *map)
...
+            dealloc_mld(interface);
+
+            wifi_interface_info_t *new_first_interface = wifi_hal_get_first_mld_interface(
+                interface);
+
+            if (new_first_interface == NULL && mld_name != NULL) {
+                nl80211_interface_enable(interface->mld_name, false);
+            }
+
+            if (old_first_interface == interface) {
```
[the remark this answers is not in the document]
bmilcz-comcast: I don't think so - deinit_bss will cause hostapd to clear the resources of the current first BSS (current hostap does no checks if there are other users in MLD group), which will trigger later free'ing of the same resources in next links (as we are in deinit procedure, they wouldn't be reinitialized), thus why in case we are removing a current first link we have to free beforehand the current non-first links.

src/wifi_hal_nl80211_utils.c
```diff
@@ -5970,3 +6006,194 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
 
     return freq;
 }
+
+static int reload_interface(wifi_interface_info_t *interface)
+{
+    char *interface_name = wifi_hal_get_interface_name(interface);
+    wifi_radio_info_t *radio = get_radio_by_rdk_index(interface->rdk_radio_index);
+
+    if (radio == NULL) {
```
mateuszCieslak-GL: radio is not used in this function at all, if i see correctly.
Depending on scenario, might be worth checking if interface != NULL  - but if its checked in caller of this function then this whole block might just be safely removed.

src/wifi_hal_nl80211_utils.c
```diff
@@ -5970,3 +6006,194 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+                return -1;
+            }
+        }
+    }
+
+    if (reload_interface(first_interface)) {
+        return -1;
+    }
+
+    if (restart_interface(first_interface) < 0) {
```
mateuszCieslak-GL: one convention please.
by numbers looks like this one should be adjusted to pure
if (restart...) ;)

bmilcz-comcast:
This PR was not merged due to ongoing issues introduced by other contributions.
This was long enough that the WAs that had to be implemented while working on link reconfig that actually this implementation took a slightly newer shape - due to this, I'm closing this PR and opening a new one for link reconfiguration.

New PR - https://github.com/rdkcentral/OneWifi/pull/1133

PR #652, opened 2026-04-02
RDKB-59818 : Enable Beacon protection for WPA3 encryption mode

Mishanok:
Reason for change: WPA3 security requires to have the properly working Beacon Protection mechanism. So, in case the MFP is Optional or Disabled, we should enable the Beacon Protection.

WPA_DRIVER_FLAGS_BEACON_PROTECTION flag required and it's enabled for certain projects for 2.4 and 5 GHz VAPs, but for 6 GHz it should always be enabled (but
    let me know if you have objections)

Also, to be able to set default BIGTK (Beacon Integrity Group Temporal Key) key, wifi HAL should implement new API for setting such keys (see this comment:
https://github.com/rdkcentral/rdk-wifi-hal/pull/648#issuecomment-4171294579)

Test Procedure:

If Device.WiFi.AccessPoint.x.Security.MFPConfig is set to Optional or Required the Beacon Frame protection should be enabled Otherwise, disabled.

To check if enabled:
1. Check that Beacon frames contain Management MIC IE
2. Check that Beacons and Probe Responses has Extended Capabilities Beacon Protection Enabled flag set to 1

Risks: High
Priority: P1

### DIFF
--- a/platform/broadcom/platform.c
+++ b/platform/broadcom/platform.c
@@ -2218,6 +2218,32 @@ static int set_ap_bss_color_value(int apIndex, uint32_t bssColor)
     return 0;
 }
 
+static void set_ap_beacon_protection(wifi_radio_info_t *radio, int apIndex)
+{
+    wifi_interface_info_t *interface;
+    struct hostapd_data *hapd;
+
+    if (radio->driver_data.extended_capa == NULL ||
+        !(radio->driver_data.capa.flags & WPA_DRIVER_FLAGS_BEACON_PROTECTION))
+        return;
+
+    interface = get_interface_by_vap_index(apIndex);
+    if (interface == NULL) {
+        wifi_hal_error_print("%s:%d: Failed to get interface for ap index: %d\n", __func__,
+            __LINE__, apIndex);
+        return;
+    }
+    hapd = &interface->u.ap.hapd;
+    wifi_hal_dbg_print("%s:%d: %s: set beacon protection %d\n", __func__, __LINE__, interface->name, hapd->conf->beacon_prot);
+
+    if (hapd->conf->beacon_prot) {
+        radio->driver_data.extended_capa[10] |= 0x10; /* Bit 84 - Beacon Protection Enabled */
+    } else {
+        radio->driver_data.extended_capa[10] &= ~0x10; /* Bit 84 - Beacon Protection Enabled */
+    }
+    v_secure_system("wl -i %s bcnprot enable %d", interface->name, hapd->conf->beacon_prot);
+}
+
 int platform_create_vap(wifi_radio_index_t r_index, wifi_vap_info_map_t *map)
 {
     wifi_hal_dbg_print("%s:%d: Enter radio index:%d\n", __func__, __LINE__, r_index);
@@ -2468,6 +2494,7 @@ int platform_create_vap(wifi_radio_index_t r_index, wifi_vap_info_map_t *map)
             wifi_setApManagementFramePowerControl(map->vap_array[index].vap_index, map->vap_array[index].u.bss_info.mgmtPowerControl);
 
             set_ap_bss_color_value(map->vap_array[index].vap_index, iconf->he_op.he_bss_color_disabled ? 0 : iconf->he_op.he_bss_color);
+            set_ap_beacon_protection(radio, map->vap_array[index].vap_index);
         } else if (map->vap_array[index].vap_mode == wifi_vap_mode_sta) {
 
             prepare_param_name(param_name, interface_name, "_akm");
@@ -4228,6 +4255,8 @@ static void platform_get_radio_caps_2g(wifi_radio_info_t *radio, wifi_interface_
     radio->driver_data.extended_capa_mask = malloc(sizeof(ext_cap));
     memcpy(radio->driver_data.extended_capa_mask, ext_cap, sizeof(ext_cap));
     radio->driver_data.extended_capa_len = sizeof(ext_cap);
+
+    radio->driver_data.capa.flags |= WPA_DRIVER_FLAGS_BEACON_PROTECTION;
 #endif // XB10_PORT || SCXER10_PORT || TCHCBRV2_PORT || SKYSR213_PORT || SCXF10_PORT
 
 // To reset the bss transition bit under extended capabilities, since its based on 2GHz vap configuration from OneWiFi.
@@ -4328,6 +4357,8 @@ static void platform_get_radio_caps_5g(wifi_radio_info_t *radio, wifi_interface_
     radio->driver_data.extended_capa_mask = malloc(sizeof(ext_cap));
     memcpy(radio->driver_data.extended_capa_mask, ext_cap, sizeof(ext_cap));
     radio->driver_data.extended_capa_len = sizeof(ext_cap);
+
+    radio->driver_data.capa.flags |= WPA_DRIVER_FLAGS_BEACON_PROTECTION;
 #endif // XB10_PORT || SCXER10_PORT || TCHCBRV2_PORT || SKYSR213_PORT
 
 // To reset the bss transition bit under extended capabilities, since its based on 5GHz vap configuration from OneWiFi.
@@ -4417,7 +4448,7 @@ static void platform_get_radio_caps_6g(wifi_radio_info_t *radio, wifi_interface_
     static const u8 eht_mcs[] = { 0x44, 0x44, 0x44, 0x44, 0x44, 0x44, 0x44, 0x44, 0x44 };
 #endif /* HOSTAPD_VERSION >= 211 */
     struct hostapd_iface *iface = &interface->u.ap.iface;
-    radio->driver_data.capa.flags |= WPA_DRIVER_FLAGS_AP_UAPSD;
+    radio->driver_data.capa.flags |= WPA_DRIVER_FLAGS_AP_UAPSD | WPA_DRIVER_FLAGS_BEACON_PROTECTION;
 
 #if defined(XB10_PORT) || defined(SCXER10_PORT) || defined(SCXF10_PORT)
     free(radio->driver_data.extended_capa);
--- a/platform/broadcom/platform.c
+++ b/platform/broadcom/platform.c
@@ -2236,11 +2236,6 @@ static void set_ap_beacon_protection(wifi_radio_info_t *radio, int apIndex)
     hapd = &interface->u.ap.hapd;
     wifi_hal_dbg_print("%s:%d: %s: set beacon protection %d\n", __func__, __LINE__, interface->name, hapd->conf->beacon_prot);
 
-    if (hapd->conf->beacon_prot) {
-        radio->driver_data.extended_capa[10] |= 0x10; /* Bit 84 - Beacon Protection Enabled */
-    } else {
-        radio->driver_data.extended_capa[10] &= ~0x10; /* Bit 84 - Beacon Protection Enabled */
-    }
     v_secure_system("wl -i %s bcnprot enable %d", interface->name, hapd->conf->beacon_prot);
 }
 
--- a/platform/broadcom/platform.c
+++ b/platform/broadcom/platform.c
@@ -2218,27 +2218,6 @@ static int set_ap_bss_color_value(int apIndex, uint32_t bssColor)
     return 0;
 }
 
-static void set_ap_beacon_protection(wifi_radio_info_t *radio, int apIndex)
-{
-    wifi_interface_info_t *interface;
-    struct hostapd_data *hapd;
-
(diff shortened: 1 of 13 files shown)


platform/broadcom/platform.c
```diff
@@ -2218,6 +2218,32 @@ static int set_ap_bss_color_value(int apIndex, uint32_t bssColor)
     return 0;
 }
 
+static void set_ap_beacon_protection(wifi_radio_info_t *radio, int apIndex)
```
bmilcz-comcast: Apart from v_secure_system call, this function seems like common code for all platforms to me.
Can't this be split into common code and platform-dependent ?

Especially since changes below are done unconditinally for all platforms - won't that cause partial configuration ?

src/wifi_hal_nl80211.c
```diff
@@ -17193,46 +17210,114 @@ int     wifi_drv_set_key(const char *ifname, void *priv, enum wpa_alg alg,
     //wifi_hal_dbg_print("%s:%d: key Info: index:%d length:%d alg:%s\n", __func__, __LINE__, key_idx, key_len, wpa_alg_to_string(alg));
     //my_print_hex_dump(key_len, key);
 
-#if HOSTAPD_VERSION < 210 //2.10
+    ret = -ENOBUFS;
+    key_msg = nlmsg_alloc();
+    if (!key_msg)
+        return ret;
+
+#if HOSTAPD_VERSION >= 210
```
bmilcz-comcast: I am fairly confident that these implementations can be made more common - adding this much macro guards will cause the code to be harder to read, please consider merging those into common implementation while retaining device-dependent behaviour.

platform/broadcom/platform.c
```diff
@@ -2218,6 +2218,32 @@ static int set_ap_bss_color_value(int apIndex, uint32_t bssColor)
...
+    if (interface == NULL) {
+        wifi_hal_error_print("%s:%d: Failed to get interface for ap index: %d\n", __func__,
+            __LINE__, apIndex);
+        return;
+    }
+    hapd = &interface->u.ap.hapd;
+    wifi_hal_dbg_print("%s:%d: %s: set beacon protection %d\n", __func__, __LINE__, interface->name, hapd->conf->beacon_prot);
+
+    if (hapd->conf->beacon_prot) {
+        radio->driver_data.extended_capa[10] |= 0x10; /* Bit 84 - Beacon Protection Enabled */
```
bmilcz-comcast: What if extended_capa is not long enough ? AFAIK, its' length varies depending on what driver passes down to us, so this should be checked before assignment.
I prefer more having defines to magic values and instead of the comment I would this to be a define instead of 0x10. AFAIK there is WLAN_EXT_CAPAB_BEACON_PROTECTION.

platform/broadcom/platform.c
bmilcz-comcast: Every code in this function (except mentioned v_secure call) - at least at a glance - looks to be common across implementations, since it enables BEACON_PROT through hostapd and not by any vendor tricks.

Your code from wifi_hal_hostapd.c enables beacon protection for all platforms, yet this code is only reachable on Broadcom (and it does seem necessary, as it enables a driver extension via a flag).

That will mean that we have enabled beacon_prot in configuration, but not in the driver on other platforms (e.g. BPi). 

Either wifi_hal_hostapd.c needs to be limited (in case it shouldn't be turned on other platforms) or this code needs to be made common.

platform/broadcom/platform.c
```diff
@@ -4228,6 +4255,8 @@ static void platform_get_radio_caps_2g(wifi_radio_info_t *radio, wifi_interface_
     radio->driver_data.extended_capa_mask = malloc(sizeof(ext_cap));
     memcpy(radio->driver_data.extended_capa_mask, ext_cap, sizeof(ext_cap));
     radio->driver_data.extended_capa_len = sizeof(ext_cap);
+
+    radio->driver_data.capa.flags |= WPA_DRIVER_FLAGS_BEACON_PROTECTION;
```
bmilcz-comcast: This unconditionally sets the flag - is this intended behaviour ? Aren't those capabilities effectively fetched from kernel ? What if this is executed for an old device not supporting this ?

src/wifi_hal_nl80211.c
```diff
@@ -17193,46 +17210,114 @@ int     wifi_drv_set_key(const char *ifname, void *priv, enum wpa_alg alg,
     //wifi_hal_dbg_print("%s:%d: key Info: index:%d length:%d alg:%s\n", __func__, __LINE__, key_idx, key_len, wpa_alg_to_string(alg));
     //my_print_hex_dump(key_len, key);
 
-#if HOSTAPD_VERSION < 210 //2.10
+    ret = -ENOBUFS;
+    key_msg = nlmsg_alloc();
```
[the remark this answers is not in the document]
bmilcz-comcast: IMO this should be aligned to codestyle that we intend to keep here, not to other project. I'm not going to be insistent on this, so in any case higher level reviewer will confirm/deny this.

src/wifi_hal_nl80211.c
bmilcz-comcast: Hmmm, frankly I don't understand. Previously, this implementation had a one guard, which just did a split between one version and the other and that was it. Right now it just feels fragmented to me.

src/wifi_hal_nl80211.c
```diff
@@ -17181,10 +17181,27 @@ int     wifi_drv_set_key(const char *ifname, void *priv, enum wpa_alg alg,
 {
     wifi_interface_info_t *interface;
     struct nl_msg *msg = NULL;
+    struct nl_msg *key_msg = NULL;
     unsigned int suite;
     int ret;
     wifi_vap_info_t *vap;
 
+#if HOSTAPD_VERSION >= 210 // 2.10
```
bmilcz-comcast: This macro is wrongly used in here.
For hostap 210 AFAIK we do not have link_ids while this way of defining the guard includes it.

src/wifi_hal_nl80211.c
```diff
@@ -17241,138 +17326,90 @@ int     wifi_drv_set_key(const char *ifname, void *priv, enum wpa_alg alg,
...
-        nla_put_u8(msg, NL80211_ATTR_MLO_LINK_ID, params->link_id);
+#if HOSTAPD_VERSION >= 210 && defined(CONFIG_GENERIC_MLO)
+    if (link_id != NL80211_DRV_LINK_ID_NA) {
+        if (nla_put_u8(msg, NL80211_ATTR_MLO_LINK_ID, link_id))
+            goto fail;
     }
 #endif // HOSTAPD_VERSION >= 211 && CONFIG_GENERIC_MLO
 
-    nla_put_u8(msg, NL80211_ATTR_KEY_IDX, params->key_idx);
-#if defined(TCXB7_PORT) || defined(TCXB8_PORT) || defined(XB10_PORT) || defined(SCXER10_PORT) || defined (TCHCBRV2_PORT) || defined(_PLATFORM_RASPBERRYPI_) || defined(_PLATFORM_BANANAPI_R4_) || defined(RDKB_ONE_WIFI_PROD)
```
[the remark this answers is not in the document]
bmilcz-comcast: well yes and no - here the guard is justified due to a change in API.
Where the pieces are common they should be made as common as possible and close to each other.

src/wifi_hal_nl80211.c
bmilcz-comcast: What if hostap is under 210 ? This way of defining the macros completely excludes older versions.

src/wifi_hal_nl80211.c
```diff
@@ -17193,46 +17210,114 @@ int     wifi_drv_set_key(const char *ifname, void *priv, enum wpa_alg alg,
...
+        msg = nl80211_drv_cmd_msg(g_wifi_hal.nl80211_id, interface, 0, NL80211_CMD_NEW_KEY);
+        if (msg == NULL) {
+            wifi_hal_dbg_print("%s:%d:Failed to allocate nl80211 message\n", __func__, __LINE__);
+            goto fail2;
+        }
+	} else if (alg == WPA_ALG_NONE && (key_flag & KEY_FLAG_RX_TX)) {
+		wifi_hal_dbg_print("%s:%d invalid key_flag to delete key", __func__, __LINE__);
+		ret = -EINVAL;
+		goto fail2;
+	} else
```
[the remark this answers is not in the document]
bmilcz-comcast: If you check the hostap version of this, it clearly has 4 conditions for the same check (assuming I am not looking at some different implementation with vendor changes)

``
if ((key_flag & KEY_FLAG_PAIRWISE_MASK) == KEY_FLAG_PAIRWISE_RX_TX_MODIFY) {
...
	} else if (alg == WPA_ALG_NONE && (key_flag & KEY_FLAG_RX_TX)) {
...
	} else if (alg == WPA_ALG_NONE) {
...
	} else {
``

platform/broadcom/platform.c
Mishanok: Made set_beacon_protection() a common thing

src/wifi_hal_nl80211.c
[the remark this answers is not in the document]
Mishanok: No more relevant

src/wifi_hal_nl80211.c
Mishanok: Reworked so only 1 ifdef guard left, as before

src/wifi_hal_nl80211.c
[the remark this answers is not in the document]
Mishanok: I proposed some solution, but it's just for the discussion, I believe. Later I will get the list of devices supporting those and will return the ifdef

platform/broadcom/platform.c
Mishanok: Removed

src/wifi_hal_nl80211.c
```diff
@@ -17193,46 +17210,114 @@ int     wifi_drv_set_key(const char *ifname, void *priv, enum wpa_alg alg,
...
 
-        types = nla_nest_start(msg, NL80211_ATTR_KEY_DEFAULT_TYPES);
-        if (!types) {
-            nl80211_nlmsg_clear(msg);
-            nlmsg_free(msg);
-        }
-        nla_put_flag(msg, NL80211_KEY_DEFAULT_TYPE_MULTICAST);
-        nla_nest_end(msg, types);
+    if (nla_put_u8(key_msg, NL80211_KEY_IDX, key_idx) ||
+        nla_put_nested(msg, NL80211_ATTR_KEY, key_msg))
```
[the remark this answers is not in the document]
Mishanok: Talked to Narendra, we aren't going to support older kernels, so we can left it as it is

platform/broadcom/platform.c
Mishanok: Not relevant. This code is not needed, I removed it. it's handled by hostapd

platform/broadcom/platform.c
```diff
@@ -4547,6 +4547,22 @@ int platform_get_reg_domain(wifi_radio_index_t radioIndex, UINT *reg_domain)
...
 
+int platform_set_beacon_prot(uint apIndex, bool isEnabled)
+{
+    wifi_interface_info_t *interface;
+
+    interface = get_interface_by_vap_index(apIndex);
+    if (interface == NULL) {
+        wifi_hal_error_print("%s:%d: Failed to get interface for ap index: %d\n", __func__,
+            __LINE__, apIndex);
+        return RETURN_ERR;
```
bmilcz-comcast, line 4565: just a suggestion but most of the code goes with either kernel compliant error codes OR with -1 as a generic error code, thus I would stay away from these RETURN_<xyz> as they are not well respected here.

src/wifi_hal_hostapd.c
```diff
@@ -637,6 +638,11 @@ int update_security_config(wifi_vap_security_t *sec, struct hostapd_bss_config *
                 case wifi_security_mode_wpa_wpa2_enterprise:
                     conf->wpa_key_mgmt |= WPA_KEY_MGMT_IEEE8021X_SHA256;
                     break;
+                case wifi_security_mode_wpa3_compatibility:
+                case wifi_security_mode_wpa3_enterprise:
+                case wifi_security_mode_wpa3_personal:
+                case wifi_security_mode_wpa3_transition:
+                    conf->beacon_prot = 1;
```
bmilcz-comcast, line 645: AFAIK there is no way to turn it on/off currently in TR-181.
Please add a device guard to make sure that only devices that in full support beacon protection have it enabled.

bmilcz-comcast:
Since this was not cleaned up and merged before Misha stopped working on this, this will be merged under
https://github.com/rdkcentral/rdk-wifi-hal/pull/761

PR #696, opened 2026-05-12
RDKBWIFI-434: [ACL_DENY] auth frame (ap -> sta) status code should be 33

sundram0711:
Introduced HAL apis to use existing hostapd apis for ACL operations.

Now hostapd responsible for the below:
-If Station is present in ACL list and try to connect to that AP,
 Then AP should drop the authentication via sending authentication frame
 with status code 33(due to insufficient bandwidth).
-If authentication allowed then AP should reject by sending assoc/re-assoc rsp
with status code 33.

Signed-off-by: Sundram Patel <sundram.p@tataelxsi.co.in>


gsathish86:
@sundram0711 
Kindly check and resolve Copilot comments.

src/wifi_hal_nl80211.c
```diff
@@ -1964,6 +1964,89 @@ static bool is_wifi_hal_rate_limit_block(unsigned short stype, mac_address_t mac
...
+    memcpy(resp.da, sta, ETH_ALEN);
+    memcpy(resp.sa, interface->mac, ETH_ALEN);
+    memcpy(resp.bssid, interface->mac, ETH_ALEN);
+
+    if (subtype == WLAN_FC_STYPE_AUTH) {
+        u16 auth_alg = le_to_host16(req->u.auth.auth_alg);
+        u16 auth_transaction = le_to_host16(req->u.auth.auth_transaction);
+
+        resp.u.auth.auth_alg = host_to_le16(auth_alg);
+        resp.u.auth.auth_transaction = host_to_le16(auth_transaction);
```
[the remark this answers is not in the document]
sundram0711: done

src/wifi_hal_nl80211.c
```diff
@@ -2082,8 +2165,8 @@ int process_frame_mgmt(wifi_interface_info_t *interface, struct ieee80211_mgmt *
         }
 #ifdef NL80211_ACL
         if (is_core_acl_drop_mgmt_frame(interface, sta)) {
-            wifi_hal_dbg_print("%s:%d: Station present in acl list dropping auth req\n",
-                                   __func__, __LINE__);
+            wifi_hal_dbg_print("%s:%d: Station present in acl list sending auth fram with rejection\n", __func__, __LINE__);
```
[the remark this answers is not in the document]
sundram0711: done

sundram0711:
> @sundram0711 Kindly check and resolve Copilot comments.

Addressed all review comments

src/wifi_hal_nl80211.c
```diff
@@ -1964,6 +1964,89 @@ static bool is_wifi_hal_rate_limit_block(unsigned short stype, mac_address_t mac
...
+    resp.u.assoc_resp.capab_info = host_to_le16(0);
+    resp.u.assoc_resp.status_code = host_to_le16(status_code);
+    resp.u.assoc_resp.aid = host_to_le16(0);
+
+    u8 *pos = (u8 *)&resp.u.assoc_resp + sizeof(resp.u.assoc_resp);
+    *pos++ = WLAN_EID_SUPP_RATES;
+    *pos++ = 1;
+    *pos++ = 0x82;
+
+    size_t len = IEEE80211_HDRLEN + sizeof(resp.u.assoc_resp) + 3;
```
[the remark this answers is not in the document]
sundram0711: removed the supported rates.

Nikita-Hakai:
Internal build: https://gerrit.teamccp.com/959311 - Success

src/wifi_hal_nl80211.c
```diff
@@ -1984,6 +1984,93 @@ static bool is_wifi_hal_rate_limit_block(unsigned short stype, mac_address_t mac
...
+
+    resp.frame_control = IEEE80211_FC(WLAN_FC_TYPE_MGMT, subtype);
+    memcpy(resp.da, sta, ETH_ALEN);
+    if (memcmp(req->da, bmac, ETH_ALEN) == 0) {
+        memcpy(resp.sa, interface->mac, ETH_ALEN);
+        memcpy(resp.bssid, interface->mac, ETH_ALEN);
+    } else {
+        memcpy(resp.sa, req->da, ETH_ALEN);
+        memcpy(resp.bssid, req->da, ETH_ALEN);
+    }
```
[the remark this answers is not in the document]
sundram0711: Copilot only suggested these changes before.

mateuszCieslak-GL (commented):
bar mutex reported by copilot, lgtm.

src/wifi_hal_nl80211.c
```diff
@@ -1983,6 +1983,114 @@ static bool is_wifi_hal_rate_limit_block(unsigned short stype, mac_address_t mac
...
+        ret = wifi_drv_send_mlme((void *)interface, (u8 *)&resp,
+            IEEE80211_HDRLEN + sizeof(resp.u.auth), 0, 0, NULL, 0);
+#endif
+
+    } else if (subtype == WLAN_FC_STYPE_ASSOC_RESP ||
+           subtype == WLAN_FC_STYPE_REASSOC_RESP) {
+
+        resp.u.assoc_resp.capab_info = host_to_le16(hostapd_own_capab_info(&interface->u.ap.hapd));
+        resp.u.assoc_resp.status_code = host_to_le16(status_code);
+        resp.u.assoc_resp.aid = host_to_le16(0);
```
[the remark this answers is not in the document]
sundram0711: addressed.

src/wifi_hal_nl80211.c
```diff
@@ -2113,8 +2230,17 @@ int process_frame_mgmt(wifi_interface_info_t *interface, struct ieee80211_mgmt *
...
+            wifi_hal_dbg_print("%s:%d: STA rejected by ACL; sending auth rejection frame\n", __func__, __LINE__);
+            if (len < IEEE80211_HDRLEN + sizeof(mgmt->u.auth)) {
+                wifi_hal_error_print("%s:%d: interface:%s auth frame too short (len=%u), cannot send rejection\n",
+                    __func__, __LINE__, interface->name, len);
+            } else {
+                ret = wifi_drv_send_mgmt_response_frame(interface, mgmt, sta, WLAN_FC_STYPE_AUTH, WLAN_STATUS_DENIED_INSUFFICIENT_BANDWIDTH);
+                if (ret < 0) {
+                    wifi_hal_error_print("%s:%d: interface:%s failed to send auth rejection frame: %d\n",
+                        __func__, __LINE__, interface->name, ret);
+                }
```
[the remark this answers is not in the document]
sundram0711: fixed.

src/wifi_hal_nl80211.c
```diff
@@ -2164,7 +2302,17 @@ int process_frame_mgmt(wifi_interface_info_t *interface, struct ieee80211_mgmt *
...
         }
-
+#ifdef NL80211_ACL
+        if (is_core_acl_drop_mgmt_frame(interface, sta)) {
+            wifi_hal_dbg_print("%s:%d: STA rejected by ACL; sending reassoc rejection response\n", __func__, __LINE__);
+            ret = wifi_drv_send_mgmt_response_frame(interface, mgmt, sta, WLAN_FC_STYPE_REASSOC_RESP, WLAN_STATUS_DENIED_INSUFFICIENT_BANDWIDTH);
+            if (ret < 0) {
+                 wifi_hal_error_print("%s:%d: interface:%s failed to send re-assoc rejection frame: %d\n",
+                     __func__, __LINE__, interface->name, ret);
+            }
```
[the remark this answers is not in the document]
sundram0711: fixed.

src/wifi_hal_nl80211.c
```diff
@@ -1983,6 +1983,120 @@ static bool is_wifi_hal_rate_limit_block(unsigned short stype, mac_address_t mac
...
+        pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+
+        resp.u.assoc_resp.capab_info = host_to_le16(capab_info);
+        resp.u.assoc_resp.status_code = host_to_le16(status_code);
+        resp.u.assoc_resp.aid = host_to_le16(0);
+
+        size_t len = IEEE80211_HDRLEN + sizeof(resp.u.assoc_resp);
+
+        wifi_hal_dbg_print("%s:%d: Sending ASSOC Response (subtype=%d, status=%d) to STA:%s\n",
+            __func__, __LINE__, subtype, status_code, to_mac_str(sta, sta_mac_str));
```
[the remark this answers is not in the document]
sundram0711: changed.

src/wifi_hal_nl80211.c
```diff
@@ -1983,6 +1983,120 @@ static bool is_wifi_hal_rate_limit_block(unsigned short stype, mac_address_t mac
     return true;
 }
 
+#ifdef NL80211_ACL
+#ifndef WLAN_STATUS_DENIED_INSUFFICIENT_BANDWIDTH
+#define WLAN_STATUS_DENIED_INSUFFICIENT_BANDWIDTH 33
+#endif
+static int wifi_drv_send_mgmt_response_frame(
```
mateuszCieslak-GL: Is it really necessary?
i.e have you tried achieving desired effect using wifi_send_response_failure() ,  wifi_send_wpa_supplicant_event(), wifi_drv_sta_deauth , wifi_hal_send_disconnect_steering_event, etc ? 
If not, please do. If yes, please tell why it failed and this approach (copying many elements of existing code) is necessary.

src/wifi_hal_nl80211.c
```diff
@@ -2113,8 +2230,17 @@ int process_frame_mgmt(wifi_interface_info_t *interface, struct ieee80211_mgmt *
...
 #ifdef NL80211_ACL
         if (is_core_acl_drop_mgmt_frame(interface, sta)) {
-            wifi_hal_dbg_print("%s:%d: Station present in acl list dropping auth req\n",
-                                   __func__, __LINE__);
+            wifi_hal_dbg_print("%s:%d: STA rejected by ACL; sending auth rejection frame\n", __func__, __LINE__);
+            if (len < IEEE80211_HDRLEN + sizeof(mgmt->u.auth)) {
+                wifi_hal_error_print("%s:%d: interface:%s auth frame too short (len=%u), cannot send rejection\n",
+                    __func__, __LINE__, interface->name, len);
+            } else {
+                ret = wifi_drv_send_mgmt_response_frame(interface, mgmt, sta, WLAN_FC_STYPE_AUTH, WLAN_STATUS_DENIED_INSUFFICIENT_BANDWIDTH);
```
narendradandu: The response is sent directly from the receive thread which can block the receive path during driver retries or aggressive blocklist reconnects, please limit this logic to Banana Pi for now.

same comment applies in all the three places

src/wifi_hal_nl80211.c
sundram0711: Okay sure, let me do the same.

sundram0711: I can see the implementation of wifi_send_response_failure() but it sends only assoc/reassoc rsp frames, 
But I also want to send the auth resp with failure.

Shall I extend this existing function to send auth rsp frame ?

src/wifi_hal_nl80211.c
sundram0711: Limiting this to Banana Pi would still leave the same concern, as the receive thread could still be blocked on Banana Pi platforms. I understand the concern around performing this operation in the receive path. If you have an alternative approach in mind that avoids blocking the receive thread while maintaining the required behavior, I'd be happy to consider it.

src/wifi_hal_nl80211.c
mateuszCieslak-GL: 1. Ive read the spec, you ought to send assoc resp with reason 33 for assoc request and auth resp with reason 33 for auth response. *NOT* send auth resp in both of this cases.

2. hal is *NOT* a place for implementing management in third party app because of convienience. Not least because reason 33 for auth/assoc req is required only within Mesh network. Non mesh client might not expect this.
As such, best place for this would  be OneWifi, or EM controler, there is already supporting infrastructure for this - you have handle_auth_Request in unified-wifi-mesh, which could be re-purposed/extended, you have ACL list management in one wifi accessible per VAP, finally you have callbacks to react to management frames in OneWifi.

The only sane way for doing this here is use as low and as simple EXISTING code as possible. So yes, extend this function to send auth rsp frame. But prefferably implement this in UnifiedWifiMesh or OneWifi.

sundram0711: Hi @mateuszCieslak-GL ,
Thanks for the feedback.

After looking more closely at the current flow, I think there may be a cleaner approach that aligns better with the existing architecture.

As I understand it, the HAL layer is currently responsible for sending and receiving 802.11 management frames to/from the driver. Received frames are forwarded to hostapd, and hostapd generates the corresponding Authentication/Association response frames and sends them back through HAL to the driver.

Before generating the response frame, hostapd already checks its internal ACL list. If a STA is denied, hostapd generates a failure response (currently using status code 1).

Instead of generating and transmitting Authentication/Association response frames directly from the HAL, my proposal is:

-Add the STA to hostapd's ACL list rather than updating a acl list of HAL for cacr.
-Update hostapd to use status code 33 (Association Denied due to Association Control policy) instead of status code 1 for stations rejected due to a Client Association Control Request.
-Let hostapd continue to generate and send the Authentication/Association response frames through the existing path.

This would allow us to reuse the existing hostapd infrastructure, avoid generating management frames in the HAL layer, and I believe it eliminate the concerns about blocking the receive thread that were raised in next comments.

If this approach sounds reasonable, I'll proceed with the required changes, perform validation/testing, and update the PR accordingly.

mateuszCieslak-GL: Yes, that approach sounds good. Though @narendradandu / @gsathish86 have the final say.
And yes, you read the architecture correctly - hal should not do much besides passing the management frames around. Thats why i mentioned OneWifi/EM agent - frames are passed up via callback. But, when no-one consumes them - they are routed to hostapd for action.
Which also seems like a good place for such hook as you noticed. Just  consider impact of such change - i.e. if you'd end up with 500+ lines patch, which will have to be migrated manually to newer versions - it might not be the best approach too.

sundram0711: Thanks for the guidance.

Since HAL already has access to the hapd context, updating the hostapd ACL list should be straightforward. From what I've seen, the implementation impact should be minimal.

hostapd already performs the ACL checks and generates the Authentication/Association responses. Therefore, instead of generating response frames in HAL, I believe we can reuse the existing hostapd flow and only update the rejection status code to 33 (as per the specification) when the STA is present in the deny list.

Hostapd already provides ACL-related APIs such as hostapd_add_acl_maclist(), hostapd_set_acl(), hostapd_ctrl_iface_acl_del_mac(), and hostapd_disassoc_deny_mac(), so I expect only minimal changes to be required in the hostpapd while reusing the existing infrastructure.

mateuszCieslak-GL:
switching to draft for the rework period.

src/wifi_hal_nl80211.c
sundram0711: @narendradandu, @gsathish86, could you please confirm if you're aligned with this direction before I proceed with the implementation and testing?

src/wifi_hal_nl80211.c
sundram0711: Added new hal api to use hostpad exsiting code, so this will resolve this issue.

src/wifi_hal_nl80211.c
sundram0711: As confirmed via email, added changes to use hostapd code for acl.

src/wifi_hal.c
```diff
@@ -2032,6 +2033,85 @@ INT wifi_hal_getScanResults(wifi_radio_index_t index, wifi_channel_t *channel, w
...
+
+    interface = get_interface_by_vap_index(apIndex);
+    if (!interface)
+        return RETURN_ERR;
+
+    if (interface->vap_info.vap_mode != wifi_vap_mode_ap)
+        return RETURN_ERR;
+
+    if (DeviceMacAddress == NULL)
+        return RETURN_ERR;
```
mateuszCieslak-GL:
```suggestion
    if ((!interface) || (interface->vap_info.vap_mode != wifi_vap_mode_ap)  || (DeviceMacAddress == NULL)
        return RETURN_ERR;
```

src/wifi_hal.c
```diff
@@ -2032,6 +2033,85 @@ INT wifi_hal_getScanResults(wifi_radio_index_t index, wifi_channel_t *channel, w
     return RETURN_OK;
 }
 
+INT wifi_hal_addHostapdDenyAclDevice(INT apIndex, CHAR *DeviceMacAddress)
+{
+    wifi_hal_info_print("%s:%d:Enter\n", __func__, __LINE__);
```
mateuszCieslak-GL:
```suggestion
    wifi_hal_dbg_print("%s:%d:Enter\n", __func__, __LINE__);
```
entry/exit logs go on lowest possible level.

src/wifi_hal.c
```diff
@@ -2032,6 +2033,85 @@ INT wifi_hal_getScanResults(wifi_radio_index_t index, wifi_channel_t *channel, w
...
+    if (!hapd->conf) {
+        pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+        return RETURN_ERR;
+    }
+
+    wifi_hal_info_print("%s:%d: CACR: Adding STAs to hostapd ACL list\n", __func__, __LINE__);
+    if (hostapd_ctrl_iface_acl_add_mac(
+            &hapd->conf->deny_mac,
+            &hapd->conf->num_deny_mac,
+            DeviceMacAddress)) {
```
mateuszCieslak-GL: roll this into max 2 lines please (there are 110-120 columns - use it :))

src/wifi_hal.c
```diff
@@ -2032,6 +2033,85 @@ INT wifi_hal_getScanResults(wifi_radio_index_t index, wifi_channel_t *channel, w
...
+    hostapd_set_acl(hapd);
+    hostapd_disassoc_deny_mac(hapd);
+
+    pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+    return RETURN_OK;
+}
+
+INT wifi_hal_delHostapdDenyAclDevice(INT apIndex, CHAR *DeviceMacAddress)
+{
+    wifi_hal_info_print("%s:%d:Enter\n", __func__, __LINE__);
```
mateuszCieslak-GL:
```suggestion
    wifi_hal_dbg_print("%s:%d:Enter\n", __func__, __LINE__);
```
as above. I would encourage to leave only middle-function log line at debug level, if youre comfortable enough its behaving as it should. If not - after some tests

src/wifi_hal.c
```diff
@@ -2032,6 +2033,85 @@ INT wifi_hal_getScanResults(wifi_radio_index_t index, wifi_channel_t *channel, w
...
+
+    interface = get_interface_by_vap_index(apIndex);
+    if (!interface)
+        return RETURN_ERR;
+
+    if (interface->vap_info.vap_mode != wifi_vap_mode_ap)
+        return RETURN_ERR;
+
+    if (DeviceMacAddress == NULL)
+        return RETURN_ERR;
```
mateuszCieslak-GL:
```suggestion
    if (!interface || (interface->vap_info.vap_mode != wifi_vap_mode_ap) || (DeviceMacAddress == NULL))
        return RETURN_ERR;
```

src/wifi_hal.c
```diff
@@ -2032,6 +2033,85 @@ INT wifi_hal_getScanResults(wifi_radio_index_t index, wifi_channel_t *channel, w
...
+    if (!hapd->conf) {
+        pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+        return RETURN_ERR;
+    }
+
+    wifi_hal_info_print("%s:%d: CACR: Removing STAs from hostapd ACL list\n", __func__, __LINE__);
+    if (hostapd_ctrl_iface_acl_del_mac(
+            &hapd->conf->deny_mac,
+            &hapd->conf->num_deny_mac,
+            DeviceMacAddress)) {
```
mateuszCieslak-GL: as above, more compact please

src/wifi_hal.c
```diff
@@ -2032,6 +2033,85 @@ INT wifi_hal_getScanResults(wifi_radio_index_t index, wifi_channel_t *channel, w
     return RETURN_OK;
 }
 
+INT wifi_hal_addHostapdDenyAclDevice(INT apIndex, CHAR *DeviceMacAddress)
```
mateuszCieslak-GL, line 2017: add 1-3 lines of comment please - what mechanism does it use, and if its hostap-version dependent. If yes (and i assume it is) - we need a macro guard for this too.

mateuszCieslak-GL (changes requested):
looks nice overall, few touches left

src/wifi_hal.c
mateuszCieslak-GL: also: Sanity check: Unless this is wired only on particular builds (i.e. easymesh enabled) , it changes reject reason for ALL devices in ALL scenarios. Have you run that past anyone? Or at least googled/checked WiFi spec it wont break anything? 

The way i see it (might be wrong though):
- hostap patch gets included only on easymesh builds (for now - bananapi)
- this code needs to be enabled only for such builds with particular hostap version. 

Ie. we need ifdef guard for hostap version AND second one for easymesh feature (or sth similar).

sundram0711: I've added comments describing the mechanism used by these APIs.

I also checked the dependency further. The implementation relies on the upstream hostapd ACL helper APIs
hostapd_ctrl_iface_acl_add_mac() and hostapd_ctrl_iface_acl_del_mac()
from ctrl_iface_ap.c / ctrl_iface_ap.h.

These APIs are part of upstream hostapd and were moved into shared files by upstream commit:

f5ac42811607 ("Move ACL control interface commands into shared files")
https://git.w1.fi/cgit/hostap/commit/?id=f5ac428116073522cc532a44900301bb9d66abdd

(March 2022)

They are also present in the hostapd 2.11 version currently used by RDK and are already used internally by hostapd and wpa_supplicant AP mode.

Based on this verification, I don't believe an additional hostap-version-specific guard is required.

For the status-code behavior, I've updated the corresponding hostap patch to apply only for EASY_MESH_NODE builds.

sundram0711: > In EasyMesh builds, ACL-based authentication/association rejections return status code 33 (WLAN_STATUS_DENIED_INSUFFICIENT_BANDWIDTH). 

> In non-EasyMesh builds, the existing behavior remains unchanged and status code 1 (WLAN_STATUS_UNSPECIFIED_FAILURE) is returned.

mateuszCieslak-GL (approved):

narendradandu (approved):

PR #697, opened 2026-05-12
RDKB-64893: Port BCAWLAN-290139 and make changes in RDKB_ONE_WIFI_PROD caused by RDKB-61540

sushmanatarajan:
BCAWLAN-290139: Change the call from iw list to reading from /sys/class/net/wlN/phy80211/name

Problem:

remap_wifi_interface_name_index_map() uses popen("iw list") to determine which kernel phy number (phy0/phy1/phy2) is assigned to each wl interface (wl0/wl1/wl2), then corrects the static interface_index_map phy_index values so that wiphy_dump_handler maps each phy to the right RDK radio index. On some boots, this function runs before the driver has published full band-capability information to nl80211, causing iw list to return empty output even though the HAL's own nl80211 socket can already see the phys. The result is a corrupted interface map, wrong phy-to-radio assignments, and HAL initialization failure.

Fix:

Replace the iw list subprocess approach with direct sysfs reads from /sys/class/net/wlN/phy80211/name. This file is created by the kernel the moment a wireless interface registers with cfg80211 — before band info is published and without requiring a subprocess. Reading it gives the kernel phy number for each wl interface directly (e.g., wl0/phy80211/name → "phy2" means wl0 belongs to kernel phy2). The parse_wiphy_band_mapping() function and its band-ID-to-phy-index translation logic are removed entirely. A retry loop (up to 30 seconds) waits for the sysfs entries to appear before proceeding; if they never appear, the remap is skipped, and the static map is used as-is.

 RDKB-61540:  RDKB-61540 added a new field in wifi_interface_name_idex_map_t  "wifi_interface_name_t mld_interface_name; /**< MLD interface name. */"
 However, the corresponding change under RDKB_ONE_WIFI_PROD  in static_interface_index_map was not done. This will cause our internal builds to fail

### DIFF
--- a/src/wifi_hal_nl80211_utils.c
+++ b/src/wifi_hal_nl80211_utils.c
@@ -331,29 +331,29 @@ static wifi_interface_name_idex_map_t static_interface_index_map[] = {
 #endif
 
 #ifdef RDKB_ONE_WIFI_PROD
-{0, 0,  "wl0.1",   "brlan0",   100,   0,      "private_ssid_2g"},
-{2, 1,  "wl1.1",   "brlan0",   100,   1,      "private_ssid_5g"},
-{0, 0,  "wl0.2",   "brlan1",   101,   2,      "iot_ssid_2g"},
-{2, 1,  "wl1.2",   "brlan1",   101,   3,      "iot_ssid_5g"},
-{0, 0,  "wl0.3",   "brlan2",   102,   4,      "hotspot_open_2g"},
-{2, 1,  "wl1.3",   "brlan3",   103,   5,      "hotspot_open_5g"},
-{0, 0,  "wl0.4",   "br106",    106,   6,      "lnf_psk_2g"},
-{2, 1,  "wl1.4",   "br106",    106,   7,      "lnf_psk_5g"},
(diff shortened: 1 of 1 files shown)


vivianecordeiro-sky:
SOlution should be only start onewifi after driver is fully open.
The new code is not generic and it is very broadcom specific.

vivianecordeiro-sky (dismissed):
Solution should be only start onewifi after driver is fully open.
The new code is not generic and it is very broadcom specific.

sushmanatarajan:
@vivianecordeiro-sky - The change is under the Broadcom specific flag. Also, this issue is seen only on specific chipsets. The change only replaces the iw list subprocess approach with direct sysfs reads from /sys/class/net/wlN/phy80211/name. This file is created by the kernel the moment a wireless interface registers with cfg80211 — before band info is published and without requiring a subprocess. From the start it should have been implemented this way instead of depending on "iw list"/ The change is already approved internally. Need to publish it here so we dont maintain patches

narendradandu (approved):

gsathish86 (approved):

PR #709, opened 2026-05-21
RDKB-65811: MLO link reconfiguration for BPi-R4

bmilcz-comcast:
Reason for change:Enable possibility of dynamically enabling/disabling MLO links for multiple MLD devices.
Test Procedure: Verify build, test MLO VAPs, enable/disable links in MLO VAPs, check if appropriate singular VAP gets reset properly, confirm connectivity with a client device.
Risks: None
Priority: P2


bmilcz-comcast:
Continuation of work from https://github.com/rdkcentral/rdk-wifi-hal/pull/583.

Depends on https://github.com/rdkcentral/OneWifi/pull/1133

platform/banana-pi/platform.c
```diff
@@ -322,33 +676,60 @@ int platform_pre_create_vap(wifi_radio_index_t index, wifi_vap_info_map_t *map)
...
             wifi_hal_error_print("%s:%d: failed to get interface for vap_index %d\n", __func__,
                 __LINE__, vap->vap_index);
-            return -1;
+            // TODO: change to return when creating interfaces via
+            // netlink is fixed
+            continue;
+        }
+
+#ifdef CONFIG_GENERIC_MLO
+        if (has_config_changed(&interface->vap_info, vap) == false) {
```
Pavlo-Uvarov:
```suggestion
        if (!has_config_changed(&interface->vap_info, vap)) {
```

platform/banana-pi/platform.c
```diff
@@ -287,18 +250,409 @@ int nvram_get_current_ssid(char *l_ssid, int vap_index)
...
+static bool wifi_hal_is_mld_link_exists(struct hostapd_data *hapd)
+{
+    struct hostapd_data *link_bss;
+
+    if (hapd->mld == NULL) {
+        return false;
+    }
+
+    dl_list_for_each(link_bss, &hapd->mld->links, struct hostapd_data, link) {
+        if (link_bss == hapd) {
```
Pavlo-Uvarov: Are you comparing pointers here?

platform/banana-pi/platform.c
```diff
@@ -287,18 +250,409 @@ int nvram_get_current_ssid(char *l_ssid, int vap_index)
...
+        wifi_hal_error_print("%s:%d: Failed to allocate memory for hostapd_mld %s\n", __func__,
+            __LINE__, interface->mld_name);
+        free(mld_unit);
+        return -1;
+    }
+
+    mld_unit->mld = mld;
+    dl_list_add_tail(&g_wifi_hal.mld_array.mld_unit, &mld_unit->mld_unit);
+
+    strncpy(mld->name, interface->mld_name, sizeof(mld->name) - 1);
```
Pavlo-Uvarov:
```suggestion
    os_strlcpy(mld->name, interface->mld_name, sizeof(mld->name));
```

platform/banana-pi/platform.c
```diff
@@ -287,18 +250,409 @@ int nvram_get_current_ssid(char *l_ssid, int vap_index)
...
+    }
+
+    hapd->mld = NULL;
+    hapd->conf->mld_ap = 0;
+    hapd->conf->okc = 0;
+
+    return 0;
+}
+
+static int teardown_mlo_vap(wifi_interface_info_t *interface)
```
Pavlo-Uvarov: Consider adding check here if that is AP VAP.

platform/banana-pi/platform.c
```diff
@@ -287,18 +250,409 @@ int nvram_get_current_ssid(char *l_ssid, int vap_index)
...
+    interface->vap_configured = false;
+
+    // Reload to update MLD
+    if (first_interface != NULL && hostapd_mld_is_first_bss(&first_interface->u.ap.hapd)) {
+        reload_vap_configuration(first_interface);
+    }
+    return 0;
+}
+
+static int setup_mlo_vap(wifi_interface_info_t *interface, wifi_vap_info_t *new_vap_config)
```
Pavlo-Uvarov: Consider adding check here if that is AP VAP.

src/wifi_hal_hostapd.c
```diff
@@ -3338,6 +3342,18 @@ int start_bss(wifi_interface_info_t *interface)
 #endif
 #endif
 #endif /* CONFIG_GENERIC_MLO */
+#ifdef CONFIG_GENERIC_MLO
```
Pavlo-Uvarov:
```suggestion
```

src/wifi_hal_nl80211.c
```diff
@@ -8836,56 +8815,76 @@ int nl80211_register_mgmt_frames(wifi_interface_info_t *interface)
...
+            if (ret == 0) {
+                wifi_hal_info_print(
+                    "%s:%d: Mgmt frames have been registered for %s frame type %d\n", __func__,
+                    __LINE__, wifi_hal_get_interface_name(interface), frame_type);
+                break;
+            } else if ((-ret) == EALREADY) {
+                wifi_hal_info_print(
+                    "%s:%d: Mgmt frames already registered for %s and frame_type %d \n", __func__,
+                    __LINE__, wifi_hal_get_interface_name(interface), frame_type);
+                usleep(wait_usec);
```
Pavlo-Uvarov, line 8920: Consider waiting in a loop while checking for the status before each wait instead of waiting for whole 500ms.

src/wifi_hal_nl80211.c
```diff
@@ -16789,26 +16816,40 @@ int nl80211_register_spurious_frames(wifi_interface_info_t *interface)
...
+        ret = execute_send_and_recv(interface->spurious_nl_cb, interface->spurious_nl_event, msg,
+            spurious_frame_register_handler, interface, NULL, NULL);
+        if (ret == 0) {
+            wifi_hal_info_print("%s:%d: Spurious frames have been registered for %s\n", __func__,
+                __LINE__, wifi_hal_get_interface_name(interface));
+            break;
+        } else if ((-ret) == EALREADY) {
+            wifi_hal_info_print("%s:%d: Spurious frame handler already registered for %s\n",
+                __func__, __LINE__, wifi_hal_get_interface_name(interface));
+            usleep(wait_usec);
```
Pavlo-Uvarov, line 16966: Consider waiting in a loop while checking for the status before each wait instead of waiting for whole 300ms.

src/wifi_hal_nl80211.c
Pavlo-Uvarov: Why do we need to wait at all in case of `-EALREADY`

src/wifi_hal_nl80211.c
Pavlo-Uvarov: Why do we need to wait at all in case of `-EALREADY`

platform/banana-pi/platform.c
```diff
@@ -287,18 +250,409 @@ int nvram_get_current_ssid(char *l_ssid, int vap_index)
...
+            return -1;
+        }
+
+        if (interface->mgmt_frames_registered) {
+            nl80211_unregister_mgmt_frames(interface);
+        }
+        if (interface->spurious_frames_registered) {
+            nl80211_unregister_spurious_frames(interface);
+        }
+        if (interface->data_frames_registered) {
```
Pavlo-Uvarov: Would it be simpler to do these checks inside called functions?
Then here you'd just call `unregister_data_frame_socket(interface);` without checking if we should call it.
It can check if flag is set inside and return from if if it's not.

platform/banana-pi/platform.c
```diff
@@ -287,18 +250,409 @@ int nvram_get_current_ssid(char *l_ssid, int vap_index)
...
+        if (interface->u.ap.hapd.conf->ssid.wpa_psk != NULL &&
+            interface->u.ap.hapd.conf->ssid.wpa_psk->next == NULL) {
+            hostapd_config_clear_wpa_psk(&interface->u.ap.hapd.conf->ssid.wpa_psk);
+        }
+        pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+
+        if (dealloc_mld(interface) != 0) {
+            wifi_hal_error_print("%s:%d: Failed to deinitialize VAP %d from MLD %s\n", __func__,
+                __LINE__, interface->vap_info.vap_index, interface->mld_name);
+            return -1;
```
Pavlo-Uvarov: I don't think we should exit here.
We should proceed with unregistration.
Or we can do unregistration before this dealloc.

platform/banana-pi/platform.c
```diff
@@ -287,18 +250,409 @@ int nvram_get_current_ssid(char *l_ssid, int vap_index)
...
+    deinit_bss(&interface->u.ap.hapd);
+    if (interface->u.ap.hapd.conf->ssid.wpa_psk != NULL &&
+        interface->u.ap.hapd.conf->ssid.wpa_psk->next == NULL) {
+        hostapd_config_clear_wpa_psk(&interface->u.ap.hapd.conf->ssid.wpa_psk);
+    }
+    pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+
+    nl80211_unregister_mgmt_frames(interface);
+    nl80211_unregister_spurious_frames(interface);
+    unregister_data_frame_socket(interface);
```
Pavlo-Uvarov: unregistering for the interface when we set up the VAP? Why?

platform/banana-pi/platform.c
bmilcz-comcast: Yes

platform/banana-pi/platform.c
bmilcz-comcast: Can you point me to where the definition of this function is ? That doesn't seem to be widely used

platform/banana-pi/platform.c
bmilcz-comcast: Consider that this is a setup_mlo_vap() which should be called only if we expect to assign this specific VAP to MLD - if this VAP is currently active as non-MLO, an interface will have these handlers present, thus why we deregister as we no longer go in "regular" non-MLO way to handle these.

All of these frames will be routed through mld0 wireless dev, which will call handler of first link in that MLD and only for this first interface, to avoid frame duplication. Later fixups to correctly dispatch it to appropriate interface is done by i.e. link_id netlink param

src/wifi_hal_hostapd.c
bmilcz-comcast: Why should this be removed ?

src/wifi_hal_nl80211.c
bmilcz-comcast: There is a comment on this case right before entering the loop. Will add the retcode to make it clear.

" // For Banana Pi in case of reconfiguration it was noticed that
   // this operation can fail as it takes time after deregistration
   // of the handlers to clean them up. Do best effort to succeed."

bmilcz-comcast: Also - consider how long it may take, I don't think that polling for this long intensively is good here.

src/wifi_hal_nl80211.c
bmilcz-comcast: As above

platform/banana-pi/platform.c
bmilcz-comcast: Actually these are already there, so I can just remove these conditionals.

platform/banana-pi/platform.c
bmilcz-comcast: Will add level above.

platform/banana-pi/platform.c
bmilcz-comcast: Valid point, will do.

platform/banana-pi/platform.c
bmilcz-comcast: Same as above - added this condition level above.

platform/banana-pi/platform.c
bmilcz-comcast: I think that current version has better visibility.

platform/banana-pi/platform.c
bmilcz-comcast: Changed this slightly - now whole teardown procedure goes through as much as possible to deallocate everything that we can. We report errors at the very end, this seems like a best tradeoff for later recovery as it may simplify all scenarios.

src/wifi_hal_nl80211.c
```diff
@@ -8836,56 +8815,76 @@ int nl80211_register_mgmt_frames(wifi_interface_info_t *interface)
...
-        if (nla_put(msg, NL80211_ATTR_FRAME_MATCH, 0, NULL) < 0) {
-            nlmsg_free(msg);
-            return -1;
-        }
+            if (nla_put(msg, NL80211_ATTR_FRAME_MATCH, 0, NULL) < 0) {
+                goto error;
+            }
+
+            ret = execute_send_and_recv(interface->nl_cb, interface->nl_event, msg,
+                mgmt_frame_register_handler, interface, NULL, NULL);
```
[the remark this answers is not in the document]
bmilcz-comcast: done

src/wifi_hal_nl80211.c
```diff
@@ -2043,9 +2043,19 @@ int process_frame_mgmt(wifi_interface_info_t *interface, struct ieee80211_mgmt *
         to_mac_str(interface->mac, interface_mac_str);
         to_mac_str(mgmt->sa, sta_mac_str);
         to_mac_str(mgmt->da, frame_da_str);
+#ifdef CONFIG_GENERIC_MLO
+        if (wifi_hal_is_mld_enabled(interface)) {
+            to_mac_str(mld_mac, mld_mac_str);
+        }
```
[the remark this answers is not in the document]
bmilcz-comcast: done

src/wifi_hal_priv.h
```diff
@@ -1469,4 +1501,11 @@ int wifi_hal_get_mac_address(const char *ifname, mac_address_t mac);
 unsigned int get_band_info_from_rdk_radio_index(unsigned int rdk_radio_index);
 int get_backhaul_sta_ifname_from_radio_index(wifi_radio_index_t index, char *ifname_out,
     size_t ifname_out_len);
+int reload_vap_configuration(wifi_interface_info_t *interface);
+int reload_interface(wifi_interface_info_t *interface);
+int restart_interface(wifi_interface_info_t *interface);
+#ifdef CONFIG_GENERIC_MLO
+int teardown_mlo_vap(wifi_interface_info_t *interface);
+int setup_mlo_vap(wifi_interface_info_t *interface, wifi_vap_info_t *new_vap_config);
+#endif
```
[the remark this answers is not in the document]
bmilcz-comcast: done

platform/banana-pi/platform.c
```diff
@@ -322,33 +333,57 @@ int platform_pre_create_vap(wifi_radio_index_t index, wifi_vap_info_map_t *map)
...
+            // TODO: change to return when creating interfaces via
+            // netlink is fixed
+            continue;
+        }
+
+#ifdef CONFIG_GENERIC_MLO
+        if (has_config_changed(&interface->vap_info, vap) == false || (interface->u.ap.conf.disable_11be == true) ||
+            (interface->vap_info.vap_mode != wifi_vap_mode_ap)) {
+            continue;
         }
```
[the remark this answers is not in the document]
bmilcz-comcast: done

src/wifi_hal_nl80211_utils.c
```diff
@@ -6082,3 +6122,554 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+
+    mld_name = wifi_hal_get_mld_name_by_interface_name(interface->name);
+    if (mld_name == NULL) {
+        wifi_hal_error_print(
+            "%s:%d: MLD interface is enabled, but interface name is unset - skipping\n", __func__,
+            __LINE__);
+        return -1;
+    }
+
+    strncpy(interface->mld_name, mld_name, sizeof(interface->mld_name) - 1);
```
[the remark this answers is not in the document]
bmilcz-comcast: done

platform/banana-pi/platform.c
```diff
@@ -322,33 +333,65 @@ int platform_pre_create_vap(wifi_radio_index_t index, wifi_vap_info_map_t *map)
...
+                return -1;
+            }
+
+            interface->vap_info.u.bss_info.mld_info.common_info.mld_enable = vap->u.bss_info.mld_info.common_info.mld_enable;
+
+            // Reload to update MLD
+            wifi_interface_info_t *first_interface = wifi_hal_get_first_mld_interface(interface);
+            if (first_interface != NULL && hostapd_mld_is_first_bss(&first_interface->u.ap.hapd)) {
+                reload_vap_configuration(first_interface);
+            }
```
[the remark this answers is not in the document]
bmilcz-comcast: 1.hapd->mld != NULL is checked internally in hostapd_mld_is_first_bss()
2.in new version this is no longer ignored.

src/wifi_hal_nl80211.c
```diff
@@ -16837,10 +16846,31 @@ int nl80211_register_bss_frames(wifi_interface_info_t *interface)
...
 #endif
 
+void nl80211_unregister_spurious_frames(wifi_interface_info_t *interface)
+{
+    if (interface->spurious_frames_registered == 0) {
+        wifi_hal_dbg_print("%s:%d: interface:%s spurious frames not registered\n", __func__,
+            __LINE__, wifi_hal_get_interface_name(interface));
+        return;
+    }
+
```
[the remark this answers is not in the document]
bmilcz-comcast: It is not needed on unregistration - since the flag is used to mark if it can be done on this particular interface, no operation will take place unless that interface really has something registered. Moreover, this is shared with non-MLO enabled code, so there is no need to complicate this further.

src/wifi_hal_nl80211.c
```diff
@@ -16950,7 +16994,32 @@ static bool is_interface_in_bridge(const char *iface, const char *bridge_name)
...
 
-static int register_data_frame_socket(wifi_interface_info_t *interface)
+void unregister_data_frame_socket(wifi_interface_info_t *interface)
+{
+    if (interface->data_frames_registered == 0) {
+        wifi_hal_dbg_print("%s:%d: interface:%s data frames not registered\n", __func__, __LINE__,
+            wifi_hal_get_interface_name(interface));
+        return;
+    }
+
```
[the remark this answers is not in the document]
bmilcz-comcast: As above.

src/wifi_hal_nl80211_utils.c
```diff
@@ -6082,3 +6122,552 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+    }
+
+    for (int i = g_wifi_hal.num_radios - 1; i >= 0; i--) {
+        wifi_interface_info_t *interface_iter = NULL;
+        wifi_radio_info_t *radio = get_radio_by_rdk_index(i);
+        if (unlikely(radio == NULL)) {
+            wifi_hal_error_print("%s:%d: interface:%s failed to get radio for index:%d\n", __func__,
+                __LINE__, interface->mld_name, interface->rdk_radio_index);
+            return -1;
+        }
```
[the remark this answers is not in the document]
bmilcz-comcast: done

src/wifi_hal_nl80211_utils.c
```diff
@@ -6082,3 +6122,552 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+{
+    unsigned int ifidx = 0;
+
+    if (nl80211_enable_ap(interface, false) < 0) {
+        wifi_hal_error_print("%s:%d: interface:%s link id:%d failed to disable AP\n", __func__,
+            __LINE__, interface->mld_name, wifi_hal_get_mld_link_id(interface));
+        return -1;
+    }
+
+    if (hostapd_mld_is_first_bss(&interface->u.ap.hapd)) {
```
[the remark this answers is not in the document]
bmilcz-comcast: hostapd_mld_is_first_bss() internally does this check.

src/wifi_hal_nl80211_utils.c
```diff
@@ -6082,3 +6122,552 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+        if (interface->spurious_frames_registered) {
+            nl80211_unregister_spurious_frames(interface);
+        }
+        if (interface->data_frames_registered) {
+            unregister_data_frame_socket(interface);
+        }
+
+        first_interface = wifi_hal_get_first_mld_interface(interface);
+        if (first_interface != NULL && hostapd_mld_is_first_bss(&first_interface->u.ap.hapd)) {
+            first_interface->vap_configured = false;
```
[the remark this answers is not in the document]
bmilcz-comcast: As above

src/wifi_hal_nl80211.c
```diff
@@ -8900,56 +8858,77 @@ int nl80211_register_mgmt_frames(wifi_interface_info_t *interface)
...
             } else {
                 wifi_hal_error_print("%s:%d: Error registering for management frames on interface "
                                      "%s error: %d (%s)\n",
                     __func__, __LINE__, wifi_hal_get_interface_name(interface), ret,
                     strerror(-ret));
-                return -1;
+                goto error;
             }
         }
     }
```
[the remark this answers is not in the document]
bmilcz-comcast: done

src/wifi_hal_nl80211.c
```diff
@@ -16899,26 +16906,40 @@ int nl80211_register_spurious_frames(wifi_interface_info_t *interface)
...
+            wifi_hal_info_print("%s:%d: Spurious frame handler already registered for %s\n",
+                __func__, __LINE__, wifi_hal_get_interface_name(interface));
+            usleep(wait_usec);
+        } else {
+            wifi_hal_error_print("%s:%d: failed to register for spurious frames on interface %s, "
+                                 "error: %d (%s)\n",
+                __func__, __LINE__, wifi_hal_get_interface_name(interface), ret, strerror(-ret));
+            goto error;
+        }
     }
```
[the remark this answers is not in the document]
bmilcz-comcast: done

src/wifi_hal_nl80211_utils.c
```diff
@@ -6082,3 +6122,554 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+
+int restart_interface(wifi_interface_info_t *interface)
+{
+    char *interface_name = wifi_hal_get_interface_name(interface);
+    wifi_radio_info_t *radio = get_radio_by_rdk_index(interface->rdk_radio_index);
+    if (unlikely(radio == NULL)) {
+        wifi_hal_error_print("%s:%d: interface:%s failed to get radio for index:%d\n", __func__,
+            __LINE__, interface_name, interface->rdk_radio_index);
+        return -1;
+    }
```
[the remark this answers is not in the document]
bmilcz-comcast: done

src/wifi_hal_nl80211_utils.c
```diff
@@ -6082,3 +6122,554 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+
+    wifi_hal_dbg_print("%s:%d: interface:%s update hostapd params\n", __func__, __LINE__,
+        interface_name);
+
+    if (update_hostap_interface_params(interface) < 0) {
+        wifi_hal_error_print("%s:%d: interface:%s failed to update hostapd params\n", __func__,
+            __LINE__, interface_name);
+        return -1;
+    }
```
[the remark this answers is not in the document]
bmilcz-comcast: done

platform/banana-pi/platform.c
```diff
@@ -322,33 +333,73 @@ int platform_pre_create_vap(wifi_radio_index_t index, wifi_vap_info_map_t *map)
...
-        vap->u.bss_info.mld_info.common_info.mld_enable =
-            interface->vap_info.u.bss_info.mld_info.common_info.mld_enable;
+
+        if (vap->u.bss_info.mld_info.common_info.mld_enable == false ||
+            vap->u.bss_info.enabled == false) {
+            if (teardown_mlo_vap(interface) != 0) {
+                wifi_hal_error_print("%s:%d: Failed to teardown link for MLD ID %d on VAP idx %d\n",
+                    __func__, __LINE__, vap->u.bss_info.mld_info.common_info.mld_id,
+                    vap->vap_index);
+                return -1;
```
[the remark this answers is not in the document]
bmilcz-comcast: If a failure is encountered, there is no way for vap map to stay consistent as this already implies that it is desynchronized with actual state. We return early to have higher layer try and set everything again, which is what would happen upon encountering the failure anyway, making all the work we did here futile.

src/wifi_hal_nl80211_utils.c
```diff
@@ -6082,3 +6122,554 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+                continue;
+            }
+
+            if (first_interface == interface_iter) {
+                continue;
+            }
+
+            if (reload_interface(interface_iter)) {
+                return -1;
+            }
```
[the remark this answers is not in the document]
bmilcz-comcast: Higher layer should take care of recovery process and reapplying the old state so that we can get up again.
HAL is meant to be an applicator and do the basic minimum of recovery to ensure a proper recover can be done at later stage.

src/wifi_hal_nl80211.c
```diff
@@ -8900,56 +8881,77 @@ int nl80211_register_mgmt_frames(wifi_interface_info_t *interface)
...
                     strerror(-ret));
-                return -1;
+                goto error;
             }
         }
     }
 
     interface->mgmt_frames_registered = 1;
-
     return 0;
```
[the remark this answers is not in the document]
bmilcz-comcast: done

src/wifi_hal_hostapd.c
bmilcz-comcast: Not applicable - had to revert this piece.

src/wifi_hal_nl80211_utils.c
```diff
@@ -6082,3 +6122,556 @@ uint16_t freq_to_primary(uint16_t freq, wifi_channelBandwidth_t chwid)
...
+        if (dealloc_mld(interface) != 0) {
+            wifi_hal_error_print("%s:%d: Failed to deinitialize VAP %d from MLD %s\n", __func__,
+                __LINE__, interface->vap_info.vap_index, interface->mld_name);
+            ret = -1;
+        }
+    }
+
+    // Necessary in case of readding
+    nl80211_remove_from_bridge(interface->name);
+
```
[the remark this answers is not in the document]
bmilcz-comcast: From my testing experience this does not work and current approach that I have does. So:
- readd MLD interface to bridge (remove + add) if VAP is being added
- readd regular interface (remove + add) if VAP is moved into "regular" mode of work (single non linked BSSID)

note that I am not removing "old" interfaces i.e. wifi0 from bridge if VAP is being moved into MLD.
Experimentally, I found that the way that currently is implemented is working.

narendradandu (approved):

PR #749, opened 2026-06-17
RDKB-59829 : support to cci external client on banana-pi and iperf

skarth597:
Reason for change: 1) banana pi as a client needs to connect to xb gateways
Test Procedure: 1) Iperf test between xb gateway and Banana pi should be successful
and the artifacts are uploaded to gateway.
Risks: Low
Priority: P1

Signed-off-by: Ramakanth Taraka <ramakanth.taraka@sky.uk>


mateuszCieslak-GL:
@skarth597 please address copilot comments.  Unless this is not valid anymore, then please close it.

PR #752, opened 2026-06-22
RDKBWIFI-472: wifi_hal: Add API stub for BTM disassociation timer control

santhoshg6:
Reason for change:
This change introduces a new HAL API wifi_setApIgnoreBTMDisassocTimer() to control whether the HAL arms the internal disassociation timer associated with BTM requests.

The API is currently implemented as a stub and returns not supported, serving as a placeholder for future platform-specific implementation.

This enables integration with the OneWifi control path without impacting existing behavior.

Dep PR: https://github.com/rdkcentral/rdkb-halif-wifi/pull/113/

### DIFF
--- a/src/wifi_hal.c
+++ b/src/wifi_hal.c
@@ -4865,7 +4865,7 @@ INT wifi_getNASta(INT apIndex, const wifi_na_sta_req_params_t *params, wifi_na_s
 
 INT wifi_setApIgnoreBTMDisassocTimer(INT apIndex, BOOL ignoreBTMDisassocTimer)
 {
-    (void)apIndex;
+    AP_INDEX_ASSERT(apIndex);
     (void)ignoreBTMDisassocTimer;
 
     wifi_hal_dbg_print("%s:%d: not supported (stub)\n", __func__, __LINE__);
--- a/src/wifi_hal.c
+++ b/src/wifi_hal.c
@@ -4870,5 +4870,5 @@ INT wifi_setApIgnoreBTMDisassocTimer(INT apIndex, BOOL ignoreBTMDisassocTimer)
 
     wifi_hal_dbg_print("%s:%d: not supported (stub)\n", __func__, __LINE__);
 
-    return RETURN_ERR;
+    return WIFI_HAL_UNSUPPORTED;
 }
--- a/src/wifi_hal.c
+++ b/src/wifi_hal.c
@@ -4863,9 +4863,9 @@ INT wifi_getNASta(INT apIndex, const wifi_na_sta_req_params_t *params, wifi_na_s
 #endif /* MXL_WIFI */
 }
 
-INT wifi_setApIgnoreBTMDisassocTimer(INT ap_index, BOOL ignoreBTMDisassocTimer)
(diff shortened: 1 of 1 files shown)


Nikita-Hakai:
Internal build: https://gerrit.teamccp.com/#/c/961022/ - success

narendradandu:
can this be done by setting the disassociation imminent bit itself from the btm request calling functions in OneWifi, this new api is introducing a vap level ignore flag which can be implemented/done through the original BTM api itself.

please provide use case and justification along with the actual implementation so that we can accept the changes.

santhoshg6:
> can this be done by setting the disassociation imminent bit itself from the btm request calling functions in OneWifi, this new api is introducing a vap level ignore flag which can be implemented/done through the original BTM api itself.
> 
> please provide use case and justification along with the actual implementation so that we can accept the changes.

The requirement is to provide a per‑VAP policy control for handling the Disassociation Imminent indication, rather than requiring every caller of wifi_setBTMRequest() to modify individual BTM requests.
Rationale for a separate API:

The setting is intended to be a persistent VAP-level configuration that applies across multiple BTM requests.
Keeping this as a HAL configuration avoids updating all OneWifi BTM request paths to manipulate the Disassociation Imminent bit.
It separates policy/configuration from the BTM request payload itself.
Platform implementations can internally suppress the Disassociation Imminent flag and associated timer based on the configured VAP policy without requiring changes in BTM request generation.

This change only adds the OneWifi → HAL-IF → HAL plumbing for the configuration. The HAL implementation is currently a stub returning WIFI_HAL_UNSUPPORTED, and platform-specific support will be implemented separately.
Use case: Some deployments require steering clients using BTM requests without triggering Disassociation Imminent behavior. A VAP-level configuration enables or disables this behavior consistently for all BTM requests originating from that VAP, rather than requiring per-request modifications.

ameydav:
I would describe the rationale differently, as it came from partners that have large field deployments:

The idea is to manipulate client behavior without causing disruptive behavior.

The original meaning of the Disassoc imminent bit and timer is to notify the client:
"Please steer to another AP, but if you don't, FYI I am going to kick you out in x seconds".

This is used for stubborn clients, that ignore you when you ask to steer through BTM, when disassoc imminent is 0.

However, our partners want a hybrid behavior:
1. Warn the client that you are going to disconnect them. This will give the client incentive to steer.
2. **Don't** disconnect them if they still ignore your BTM. This will prevent them from being disconnected when they choose not to move.

This set of PRs is intended to accommodate this case.

ameydav:
We need a fix on top of https://github.com/rdkcentral/rdkb-halif-wifi/pull/113 (I no longer have ability to comment there, I guess because it was merged).

The bug is not in the code, but in the C comment describing the code.

```
* When set to true, the HAL will **not set the Disassociation Imminent flag** in outgoing
* BTM Request frames for this VAP and will not arm the corresponding disassociation timer.
* The setting applies only to the specified VAP identified by apIndex.

```
The comment in bold is wrong, and might explain why the Github automated review asked us to change the function name.
We are not removing the Disassociation Imminent flag. 
We are removing the internal timer that causes disassociation, even when this Disassociation Imminent flag is set to 1.

ameydav (approved):

mateuszCieslak-GL (approved):

PR #761, opened 2026-06-24
RDKB-59818 : Enable Beacon protection for WPA3 encryption mode

bmilcz-comcast:
Reason for change: WPA3 security requires to have the properly working Beacon Protection mechanism. So, in case the MFP is Optional or Disabled, we should enable the Beacon Protection.

WPA_DRIVER_FLAGS_BEACON_PROTECTION flag required and it's enabled for certain projects for 2.4 and 5 GHz VAPs, but for 6 GHz it should always be enabled (but
    let me know if you have objections)

Also, to be able to set default BIGTK (Beacon Integrity Group Temporal Key) key, wifi HAL should implement new API for setting such keys (see this comment:
https://github.com/rdkcentral/rdk-wifi-hal/pull/648#issuecomment-4171294579)
This was developed under:
https://github.com/rdkcentral/rdk-wifi-hal/pull/652

Test Procedure:

If Device.WiFi.AccessPoint.x.Security.MFPConfig is set to Optional or Required the Beacon Frame protection should be enabled Otherwise, disabled.

To check if enabled:
1. Check that Beacon frames contain Management MIC IE
2. Check that Beacons and Probe Responses has Extended Capabilities Beacon Protection Enabled flag set to 1

Beacon Protection is explicitly turned off for now until a requirement for it appears and/or there is ability to set it on/off on demand.

Risks: High
Priority: P1

### DIFF
--- a/src/wifi_hal_hostapd.c
+++ b/src/wifi_hal_hostapd.c
@@ -618,6 +618,7 @@ int update_security_config(wifi_vap_security_t *sec, struct hostapd_bss_config *
 
 #ifdef CONFIG_IEEE80211W
     conf->ieee80211w = (enum mfp_options)sec->mfp;
+    conf->beacon_prot = 0;
     switch (conf->ieee80211w) {
         case MGMT_FRAME_PROTECTION_REQUIRED:
             conf->wpa_key_mgmt &= ~(WPA_KEY_MGMT_PSK | WPA_KEY_MGMT_IEEE8021X);
@@ -637,6 +638,14 @@ int update_security_config(wifi_vap_security_t *sec, struct hostapd_bss_config *
                 case wifi_security_mode_wpa_wpa2_enterprise:
                     conf->wpa_key_mgmt |= WPA_KEY_MGMT_IEEE8021X_SHA256;
                     break;
+#ifdef BEACON_PROT
+                case wifi_security_mode_wpa3_compatibility:
+                case wifi_security_mode_wpa3_enterprise:
+                case wifi_security_mode_wpa3_personal:
+                case wifi_security_mode_wpa3_transition:
+                    conf->beacon_prot = 1;
+                    break;
+#endif
                 default:
                     break;
             }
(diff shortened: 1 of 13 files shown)


narendradandu (dismissed):

src/wifi_hal_hostapd.c
```diff
@@ -637,6 +638,13 @@ int update_security_config(wifi_vap_security_t *sec, struct hostapd_bss_config *
                 case wifi_security_mode_wpa_wpa2_enterprise:
                     conf->wpa_key_mgmt |= WPA_KEY_MGMT_IEEE8021X_SHA256;
                     break;
+#ifdef BEACON_PROT
+                case wifi_security_mode_wpa3_compatibility:
+                case wifi_security_mode_wpa3_enterprise:
+                case wifi_security_mode_wpa3_personal:
+                case wifi_security_mode_wpa3_transition:
+                    conf->beacon_prot = 1;
+#endif
```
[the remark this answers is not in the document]
bmilcz-comcast: Added missing break.

src/wifi_hal_priv.h
```diff
@@ -860,6 +861,7 @@ typedef struct {
     platform_set_dfs_t                platform_set_dfs_fn;
     platform_get_radio_caps_t         platform_get_radio_caps_fn;
     platform_get_RegDomain_t platform_get_RegDomain_fn;
+    platform_set_beacon_prot_t platform_set_beacon_prot_fn;
```
mateuszCieslak-GL, line 864: please align this and line above with previous ones ;)

src/wifi_hal_priv.h
```diff
@@ -1417,6 +1420,7 @@ platform_set_offload_mode_t         get_platform_set_offload_mode_fn();
 platform_set_dfs_t                  get_platform_dfs_set_fn();
 platform_get_radio_caps_t           get_platform_get_radio_caps_fn();
 platform_get_RegDomain_t get_platform_get_RegDomain_fn();
+platform_set_beacon_prot_t get_platform_set_beacon_prot_fn();
```
mateuszCieslak-GL, line 1423: please align this and line above with previous ones ;)

narendradandu (approved):

PR #766, opened 2026-06-25
DFS Functionality support for Banana-Pi in RDK-WIFI-HAL

prathbr:
1. Platform Enablement for BANANA_PI_PORT
2. DFS Channel Handling Enhancements
3. BananaPi Specific Channel Switch behaviour
4. DFS CAC Completion Logic Refinement
5. CSA Cleanup on BananaPi
6. Regulatory and Channel Definition Updates

Impact
• Enables stable DFS and channel management on BananaPi platforms. • Prevents unnecessary channel reconfiguration and hostapd CSA issues. • Maintains compatibility with CMXB7.
• Improves regulatory compliance and channel coverage.


src/wifi_hal_nl80211_events.c
```diff
@@ -1803,10 +1819,33 @@ int process_global_nl80211_event(struct nl_msg *msg, void *arg)
     case NL80211_CMD_RADAR_DETECT:
         // To handle CAC Finish and CAC Abort for DFS. These event involve only the primary
         // interface of the radio.
-        if (!(event_type == NL80211_RADAR_CAC_FINISHED || event_type == NL80211_RADAR_CAC_ABORTED))
-            break;
-
-        /* fall through */
+        if (event_type == NL80211_RADAR_CAC_FINISHED || event_type == NL80211_RADAR_CAC_ABORTED || event_type == NL80211_RADAR_CAC_STARTED) {
+
```
bmilcz-comcast, line 1833: This code will be executed for other platforms as well - I don't think this is desirable, since DFS i.e. on Broadcom is handled in some form in separate calls to external binary

src/wifi_hal_nl80211.c
```diff
@@ -19469,16 +19474,22 @@ int nl80211_dfs_radar_cac_finished(wifi_interface_info_t *interface, int freq, i
     radio_param.channel = channel;
     radio_param.channelWidth = bandwidth;
 
-    if( wifi_hal_setRadioOperatingParameters(interface->vap_info.radio_index, &radio_param) ) {
-        wifi_hal_error_print("nl80211-%s:%d wifi_hal_setRadioOperatingParameters Failed \n", __func__, __LINE__);
-    }
-    pthread_mutex_lock(&g_wifi_hal.hapd_lock);
-    interface->u.ap.iface.cac_started = 0;
-    pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+#ifdef BANANA_PI_PORT
```
bmilcz-comcast: Why this is only for BananaPi and not for CMXB7 ?

src/wifi_hal_nl80211_events.c
```diff
@@ -1179,6 +1179,22 @@ static void nl80211_ch_switch_notify_event(wifi_interface_info_t *interface, str
...
 
+#ifdef BANANA_PI_PORT
+    if ((wifi_chan_event_type == WIFI_EVENT_DFS_RADAR_DETECTED) && ((radio_channel_param.sub_event == WIFI_EVENT_RADAR_CAC_FINISHED) || (radio_channel_param.sub_event == WIFI_EVENT_RADAR_CAC_ABORTED))) {
+        if (interface->vap_info.vap_mode == wifi_vap_mode_ap && interface->u.ap.hapd.csa_in_progress) {
+            pthread_mutex_lock(&g_wifi_hal.hapd_lock);
+            wifi_interface_info_t *dfs_interface;
+            hash_map_foreach(radio->interface_map, dfs_interface) {
+                if (dfs_interface->u.ap.hapd.csa_in_progress) {
+                    wifi_hal_info_print("%s:%d: ******** csa_in_progress = %d\n", __func__, __LINE__, interface->u.ap.hapd.csa_in_progress);
+                    hostapd_cleanup_cs_params(&dfs_interface->u.ap.hapd);
```
bmilcz-comcast, line 1199: I'm having some issues in understanding this. All current platforms handle these events and AFAIK nl80211_ch_switch_notify_event does this. Why do we have to manually clear the cs params in here ?

prathbr: nl80211_switch_channel() calls hostapd_switch_channel(&interface->u.ap.hapd, &csa_settings).
hostapd_switch_channel() rejects a new channel switch if [interface->u.ap.hapd.csa_in_progress] is still true.
On Banana Pi, we observed hostapd still retained stale CSA state and failed to switch channel. 
That leaves csa_in_progress=true even though the DFS flow is done. 

nl80211_ch_switch_notify_event() does cleanup csa_in_progress  if (wifi_chan_event_type == WIFI_EVENT_CHANNELS_CHANGED) 
But it doesn't cleanup csa_in_progress if (wifi_chan_event_type == WIFI_EVENT_DFS_RADAR_DETECTED)

src/wifi_hal_nl80211_events.c
prathbr: Changed this to execute only for BANANA_PI_PORT

src/wifi_hal_nl80211.c
prathbr: Made this common for all platforms

src/wifi_hal_nl80211_events.c
bmilcz-comcast: Failure to switch the channel indicates a problem (or ongoing CSA during the requst).
For BananaPi + EasyMesh, cs_count is quite high (20). Was there no event for some interface that switch was done ?
If this is MLO then maybe we lack proper handling of this specific case, as the procedure for MLO is quite different (we need to notify partner links as well)

src/wifi_hal_nl80211_events.c
bmilcz-comcast: Correct me if I'm wrong, but is there even a primary interface on BananaPi ? Aren't they essentially private VAPs ?

platform/banana-pi/platform.c
```diff
@@ -545,9 +512,34 @@ int platform_get_radio_phytemperature(wifi_radio_index_t index,
     return 0;
 }
 
+int wifi_setRadioDfsBootupEnable(wifi_radio_index_t index, bool enable)
```
anandhkr-plume, line 515: Could see wifi_setRadioDfsAtBootupEnable defined in other platform.c file

src/wifi_hal.c
anandhkr-plume, line 948: In case of Banana Pi, how is CAC getting triggered? As far as I remember nl80211_switch_channel will only trigger channel change. 
If hostap/netlink are handling DFS, this function will trigger CAC.

src/wifi_hal_nl80211_events.c
```diff
@@ -1882,12 +1826,37 @@ int process_global_nl80211_event(struct nl_msg *msg, void *arg)
...
+
+            // Map VAP interface to primary interface for DFS events
+            if (interface != NULL) {
+                wifi_radio_info_t *radio = get_radio_by_rdk_index(interface->vap_info.radio_index);
+                if (radio != NULL) {
+                    wifi_interface_info_t *mgt_interface;
+                    if (g_wifi_hal.platform_flags & PLATFORM_FLAGS_UPDATE_WIPHY_ON_PRIMARY) {
+                        mgt_interface = get_primary_interface(radio);
+                    } else {
+                        mgt_interface = get_private_vap_interface(radio);
```
anandhkr-plume, line 1842: Function **nl80211_dfs_radar_event** has similar implementation, it will handle this case. Please check if this code is needed

src/wifi_hal_nl80211_events.c
anandhkr-plume: Reason might be due to CAC not started before Channel switch.
Ideally in case of DFS Channels, CAC should be triggered and finished before switching the channel.

If CAC/Radar Events are handled by OneWifi, CAC should be started by this function nl80211_start_dfs_cac. Once CAC is done, netlink will send CAC Finish event to hal which will call nl80211_dfs_radar_cac_finished.

nl80211_dfs_radar_cac_finished will proceed to set the channel in the netlink/driver.

Note: If the DFS channel is not set under 2 seconds after CAC Finish is received, CAC will be expired

src/wifi_hal_nl80211.c
```diff
@@ -19391,7 +19386,11 @@ int nl80211_dfs_cac_started(wifi_interface_info_t *interface, int freq, int ht_e
             interface->vap_info.radio_index);
         return 0;
     }
-
+#ifdef BANANA_PI_PORT
+    pthread_mutex_lock(&g_wifi_hal.hapd_lock);
+    interface->u.ap.iface.cac_started = 1;
+    pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+#endif
```
anandhkr-plume, line 19393: cac_started is set as 1 in nl80211_start_dfs_cac once CAC is started

anandhkr-plume (changes requested):

src/wifi_hal_nl80211_events.c
prathbr: We did not use nl80211_start_dfs_cac() because, in the flow we are following, nl80211_switch_channel() is already the path that takes care of the channel change. We observed that the normal channel-switch path clears csa_in_progress correctly, but in the DFS-related path that cleanup was not happening. As a result, hostapd could retain stale CSA state after the DFS event, and a subsequent switch could be rejected because csa_in_progress was still set. That is why the change is scoped to clearing that stale state when a new switch is required after radar detection or CAC abort. We may have to remove this fix for cac finished event and do this for radar detected and cac aborted.

anandhkr-plume: Does invoking nl80211_switch_channel() directly trigger CAC and bring the radio down for 1 min?
Are the channels set properly in the Radio if we use nl80211_switch_channel() to switch to a DFS Channel?

Does the Driver handle DFS events like CAC Start, CAC Finish when a DFS Channel is set?

prathbr: Does invoking nl80211_switch_channel() directly trigger CAC and bring the radio down for 1 min?
Yes it does. 
Are the channels set properly in the Radio if we use nl80211_switch_channel() to switch to a DFS Channel?
Yes channel is set properly in the radio.

Does the Driver handle DFS events like CAC Start, CAC Finish when a DFS Channel is set?
Yes driver handles DFS operations and provide events to HAL after cac_started, cac_finished, cac_aborted, nop_finished, radar_detected

anandhkr-plume: Okay. Got it

src/wifi_hal_nl80211.c
anandhkr-plume, line 19587: In case of Radar detection, is wifi_hal_setRadioOperatingParameters function getting called?
This function should be invoked, to set Non DFS Channel.

If Driver is handling channel change in case of Radar Detection, ideally we have to update the channel to hostap and inform Onewifi about channel change

src/wifi_hal_nl80211_events.c
anandhkr-plume: > Does invoking nl80211_switch_channel() directly trigger CAC and bring the radio down for 1 min? Yes it does. Are the channels set properly in the Radio if we use nl80211_switch_channel() to switch to a DFS Channel? Yes channel is set properly in the radio.
> 
> Does the Driver handle DFS events like CAC Start, CAC Finish when a DFS Channel is set? Yes driver handles DFS operations and provide events to HAL after cac_started, cac_finished, cac_aborted, nop_finished, radar_detected

@prathbr 
If Driver is handling all DFS Events, and if it only notifies/informs Onewifi or Hal about the channel change, I think we dont need to call nl80211_dfs_* functions in wifi_hal_nl80211.c.

This notification/update will done by nl80211_ch_switch_notify_event.

nl80211_dfs_* functions in wifi_hal_nl80211.c are invoked to handle DFS Events by Onewifi/Hal

src/wifi_hal.c
bmilcz-comcast: You are correct.

For BananaPi, channel switching only triggers CSA. For BananaPi, this may be necessary to add a define in here for CAC to be started.

src/wifi_hal_nl80211_events.c
bmilcz-comcast: This is not going to work for MLO-enabled versions of BPi as well. Is this feature focused only on delivering DFS for non-MLO APs ?

src/wifi_hal_nl80211_events.c
bmilcz-comcast: >   DFS-related path that cleanup was not happening
To me this still points to some gaps in alignment for BPi. Especially that in above comment from @anandhkr-plume for wifi_hal_setRadioOperatingParameters for XB7 there is no addition of BananaPi port that is supposed to follow the same path as XB7 (as judged by how that support was extended in other places)

src/wifi_hal_nl80211_events.c
anandhkr-plume: Yes, it was developed for non-MLO APs. Not sure about MLO APs

src/wifi_hal_nl80211_events.c
```diff
@@ -1882,12 +1802,13 @@ int process_global_nl80211_event(struct nl_msg *msg, void *arg)
     interface = get_interface_by_if_index(ifidx, link_id);
     switch (gnlh->cmd) {
     case NL80211_CMD_RADAR_DETECT:
-        // To handle CAC Finish and CAC Abort for DFS. These event involve only the primary
-        // interface of the radio.
-        if (!(event_type == NL80211_RADAR_CAC_FINISHED || event_type == NL80211_RADAR_CAC_ABORTED))
-            break;
+           // To handle CAC Finish and CAC Abort for DFS. These event involve only the primary
+           // interface of the radio.
```
bmilcz-comcast: If no changes then remove the file from the commit.

From what I see as well you have removed your mapping changes - I've tested your previous PR with some of my fixes and I also think it was not necessary. Still need to review above other DFS scenarios (abort/pre-CAC etc.)

src/wifi_hal_nl80211.c
```diff
@@ -19619,15 +19503,17 @@ int nl80211_dfs_radar_cac_finished(wifi_interface_info_t *interface, int freq, i
     radio_param.channel = channel;
     radio_param.channelWidth = bandwidth;
 
-    if( wifi_hal_setRadioOperatingParameters(interface->vap_info.radio_index, &radio_param) ) {
-        wifi_hal_error_print("nl80211-%s:%d wifi_hal_setRadioOperatingParameters Failed \n", __func__, __LINE__);
-    }
-    pthread_mutex_lock(&g_wifi_hal.hapd_lock);
-    interface->u.ap.iface.cac_started = 0;
-    pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+    if (interface->u.ap.iface.cac_started) {
```
bmilcz-comcast: What exactly this condition protect us from ? Is this called for interfaces not partaking in CAC ?

src/wifi_hal.c
bmilcz-comcast, line 944: Missing define here

bmilcz-comcast (commented):
I am having issues with building this out-of-the-box - I had to fix some of the errors in order to test this

src/wifi_hal_nl80211.c:19060:37: error: 'DFS_NON_DFS_ONLY' undeclared
src/wifi_hal_nl80211.c:19367:52: error: variable 'ht_enabled' set but not used
src/wifi_hal_nl80211.c:19367:43: error: unused variable 'cf1'

I am building this with CONFIG_GENERIC_MLO (so generic_mlo DISTRO_FEAT enabled), which ups hostapd version to 2.11.

After running this, encountered issues due to -ENETDOWN coming from nl80211_start_dfs_cac disabling interface through nl80211_interface_enable. If we do not bring the wdev down while issuing the command, no errors are seen.

There are also some issues with pre-CAC as well (improper interface pointed)

src/wifi_hal_nl80211.c
```diff
@@ -19596,7 +19452,35 @@ int nl80211_dfs_radar_cac_aborted(wifi_interface_info_t *interface, int freq, in
     if (update_channel_flags() != 0) {
         wifi_hal_error_print("%s:%d update_channel_flags failed \n", __func__, __LINE__);
     }
+#endif
+
+#ifdef BANANA_PI_PORT
```
bmilcz-comcast, line 19457: If we follow the same path for DFS (basically utilizing hostapd to do this for us) when why here we have a completely different handling and we bypass above code completely (missing BANANAPI define) ?

src/wifi_hal_nl80211.c
bmilcz-comcast, line 19206: missing define

src/wifi_hal_nl80211.c
```diff
@@ -19596,7 +19452,35 @@ int nl80211_dfs_radar_cac_aborted(wifi_interface_info_t *interface, int freq, in
...
+    pthread_mutex_lock(&g_wifi_hal.hapd_lock);
+    interface->u.ap.iface.cac_started = 0;
+    pthread_mutex_unlock(&g_wifi_hal.hapd_lock);
+
+    if (interface->vap_info.vap_mode == wifi_vap_mode_ap) {
+        pthread_mutex_lock(&g_wifi_hal.hapd_lock);
+        wifi_interface_info_t *dfs_interface;
+        hash_map_foreach(radio->interface_map, dfs_interface) {
+            if (dfs_interface->u.ap.hapd.csa_in_progress) {
+                wifi_hal_info_print("%s:%d: CAC is aborted,clearing stale CSA params for %s csa_in_progress=%d\n",
```
bmilcz-comcast, line 19476: when I merged this loop into code above, right before wifi_hal_setRadioOperatingParameters I have not noticed any of those stale prints which would indicate there is a stale csa_in_progress.

How are you triggering this exactly ?

src/wifi_hal.c
podongre: For BANANA_PI, we are using the nl80211_dfs_cac_started callback to handle the CAC start event and setting cac_started = 1 in the same callback

src/wifi_hal_nl80211.c
podongre: We tried adding the BANANA_PI handling to nl80211_start_dfs_cac() and starting CAC through this function. However, when CAC was started on channel 52, we observed the following errors:
 
nl80211_enable_ap:7623: Stopping ap on interface: 25
wifi_drv_flush:13533: nl80211: flush -> DEL_STATION wifi1 (all)
wifi_drv_sta_deauth:13290: Enter ff:ff:ff:ff:ff:ff 3
wifi_sta_remove:12054: interface:wifi1 send deauth from:02:02:20:25:41:3b to:ff:ff:ff:ff:ff:ff reason:3
error_handler:3579: kernel error: -2
wifi_sta_remove:12075: failed to sent deauth/disassoc, error:-2 (No such file or directory)
...
nl80211_start_dfs_cac:18216 interface:wifi1 Disabled for CAC
nl80211_enable_ap:7623: Stopping ap on interface: 25
error_handler:3579: kernel error: -100
nl80211_enable_ap:7626: Error stopping/starting ap: -100 (Network is down)
 
Based on this behavior, we are continuing to use the existing nl80211_dfs_cac_started callback for BANANA_PI rather than changing the CAC-start flow through nl80211_start_dfs_cac().

src/wifi_hal.c
podongre: This code path invokes nl80211_start_dfs_cac(). We attempted to enable the missing definition for the BananaPi platform but encountered the issues described in the other review comment.
As an alternative, the BananaPi flow currently handles this through the nl80211_dfs_cac_started callback, where the corresponding flag is set once CAC initiation is confirmed. Therefore, the definition is intentionally not enabled here.

src/wifi_hal_nl80211.c
podongre: The CSA state is typically cleared upon receiving NL80211_CMD_CH_SWITCH_NOTIFY, and this path works as expected when hostapd successfully completes the channel switch.
However, on BananaPi, when a radar event results in CAC_ABORTED, the driver does not generate a corresponding NL80211_CMD_CH_SWITCH_NOTIFY event. As a result, the existing cleanup path is never executed and the CSA state remains set.
The additional handling here is intended specifically for that scenario. During CAC_ABORTED processing, We explicitly clears the CSA state to ensure proper cleanup when the switch notification is not received. This is not intended as an alternative channel-switch flow, but rather as a platform-specific recovery path for the missing notification case.

src/wifi_hal_nl80211.c
podongre: Yes, the condition ensures that the CAC-finished handling is performed only for a primary interface that actually had CAC started.

src/wifi_hal_nl80211.c
bmilcz-comcast: On the first error:
I don't think this is relevant - this is an issue in kicking out STA out of that specific BSS (and it is not any particular STA, but a broadcast/wildcard address) - this can be seen even during normal operation (i.e. when bringing the VAP down or switching link during MLO). Although it should be fixed in future, this is not a blocker for our task.

Second error:
Seems to stem from code in nl80211_start_dfs_cac() - this function apart from tearing down BSS sets wireless interface off via nl80211_interface_enable(), thus why this error (and following hostapd_handle_dfs failure) is present. When this is not performed for Banana Pi, this handling goes through as expected (i.e. commented out or switched off via #ifndef BANANA_PI_PORT - I've posted about this WA in another comment where I mentioned -ENETDOWN).

This might work for regular cases but I do see a problem in current path if there would come another request for channel change while CAC is active. Basically I do not see it being restarted for another frequency which is what should happen if there was no background check being run.

Keeping this as is would create another way to enable DFS, where we partially share some logic with some other platform and further fragment rdk-wifi-hal, which is what I would like to avoid.

Given this, I would prefer to keep a uniform approach with CMXB7 - which is doable with slight adjustments to this 'Network is down' (-ENETDOWN) issue as I described, aligning the remainder to CMXB7 and getting rid of #ifndef under set_freq_and_interface_enable.

If needed, we can synchronize on the approach for this where I can suggest the changes or we can work our way in scope of this PR further.

src/wifi_hal_nl80211.c
bmilcz-comcast: I have checked on openWRT for clean reference and on netlink capture I do see NL80211_CMD_CH_SWITCH_NOTIFY accompanying the CAC_ABORTED events - both in case where 2nd request cancels CAC and when software radar simulation is done (although in this case, it was sent to non-DFS channel that was different than the initial one). Was this confirmed by netlink captures ?

src/wifi_hal_nl80211.c
bmilcz-comcast: Actually I would move this check at top of the function to do exit-early in case if cac is not started for interface

src/wifi_hal.c
bmilcz-comcast, line 944: Bot has improperly resolved this conversation. Please refer to my other comments on unification of this approach.

src/wifi_hal_nl80211.c
podongre: The stale CSA parameter cleanup from nl80211_dfs_radar_cac_aborted() is handled only for the BANANA_PI-specific case, where the main event is WIFI_EVENT_DFS_RADAR_DETECTED and the subevent is NL80211_RADAR_CAC_ABORTED.

This cleanup is triggered when CAC is in progress on a DFS channel and a radar is detected on the same channel, causing the ongoing CAC to be aborted.

platform/banana-pi/platform.c
podongre: Renamed the function from wifi_setRadioDfsBootupEnable to wifi_setRadioDfsAtBootupEnable.

bmilcz-comcast:
Since some time has passed, I found a little bit of time to put some things together based on this PR which includes the direction that I would like to follow. You can take a look at my current work and use this in your changes

https://github.com/bmilcz-comcast/rdk-wifi-hal/tree/dfs-fix

PR #790, opened 2026-07-20
TCXB8-4143: Improve wifi_hal_init() resiliency on boot.

mateuszCieslak-GL:
Reason for change : OneWifi crashes with SIGABRT via assert (hal_initialized == RETURN_OK), raised because wifi_hal_init() failed before nl_recv_func creation. The most likely source is nl80211_init_primary_interfaces(): wifi_hal_init() busy-waits for the wl *module* to be Live, but takes exactly one shot at enumerating the primary wlX *netdevs*, which the driver registers asynchronously only after the module flips to Live.
Proposed Fix : When the primary interface of a present radio is not found, re-run the per-wiphy NL80211_CMD_GET_INTERFACE in a bounded poll loop before failing. On healthy boots the first lookup succeeds and no latency is added.  Also fixes the sign bug in the SET_INTERFACE error path, which made fail-fast branch dead code.
Test Procedure : Crash should not occur. On previously failed builds log should appear: "primary interface of dev:N appeared after X ms".
Priority: P0
Risks: Low

### DIFF
--- a/src/wifi_hal_nl80211.c
+++ b/src/wifi_hal_nl80211.c
@@ -6191,9 +6191,52 @@ int nl80211_delete_interfaces(wifi_radio_info_t *radio)
     return 0;
 }
 
+/* TCXB8-4143: the wl driver registers the primary wlX netdevs asynchronously
+ * after its module reports "Live" in /proc/modules, so the wait at the lsmod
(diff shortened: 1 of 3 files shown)


mateuszCieslak-GL:
Not needed, at least for now.

PR #798, opened 2026-07-24
SHARMAN-4192: Eco mode is not getting enabled (DSPS SDK 6.3)

sureskps:
Reason for change:
- Fix eco mode commands for DSPS SDK 6.3: use dsps power_down/power_up instead of deprecated edpddn/edpdup.
- Refresh phy index and nl80211 interface ifindices after PCIe cycle on eco mode power-up
- Fix set_interface_properties to match by interface name only Test Procedure: Enable and disable eco mode

Risks: low
Priority: P0

### DIFF
--- a/src/wifi_hal_nl80211_utils.c
+++ b/src/wifi_hal_nl80211_utils.c
@@ -2204,8 +2204,14 @@ int set_interface_properties(unsigned int phy_index, wifi_interface_info_t *inte
     /* Set interface properties for VAP interfaces */
     for (i = 0; i < get_sizeof_interfaces_index_map(); i++) {
         map = &interface_index_map[i];
+#if defined(SCXER10_PORT) || defined(SKYSR213_PORT)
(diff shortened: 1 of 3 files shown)


src/wifi_hal_nl80211_utils.c
```diff
@@ -2201,11 +2201,12 @@ int set_interface_properties(unsigned int phy_index, wifi_interface_info_t *inte
 
     vap->vap_index = 0;
 
-    /* Set interface properties for VAP interfaces */
+    /* Set interface properties for VAP interfaces.
+     * Match by interface name only: phy_index is not checked because the
+     * phy/wiphy index can change after a PCIe power cycle (eco mode power-up)*/
```
mateuszCieslak-GL: I was expecting this to be under #ifdef. Have you performed any tests on different devices and they behave as expected after this change?

sureskps: Updated.

PR #807, opened 2026-08-03
RDKBWIFI-541: Report global operating class, not country specific index

dkyncu:
get_op_class_from_radio_params() returned the country table's own index (EU 4, US 12 for 2.4GHz 20MHz) instead of the IEEE global operating class (81). Consumers expect Table E-4 values: the radio oper params feed EasyMesh/DataElements CurrentOperatingClassProfile.Class, and RRM beacon reports carry the value over the air. The bug was masked on 5GHz where country and global classes coincide, and the 6GHz/global fallback paths of the same function already return global classes. In tree consumers that interpret the value (ieee80211_chan_to_freq, country_to_global_op_class) accept global input unchanged, so behavior elsewhere is identical.


dkyncu:
Hi @narendradandu

Could you please review this?

dkyncu:
Hi @mateuszCieslak-GL @narendradandu 

Could you please review this?

mateuszCieslak-GL (approved):

